PR #254, opened 2023-01-05
chore(workflows): include tag job in release

williamboman:
Triggering other workflows from within a workflow is
difficult/impossible, so couple these together.


williambotman:
@mason-org/triage

PR #277, opened 2023-01-07
feat: add ability to override source definition for historic version ranges

williamboman:
This mechanism applies to all sources, not just GitHub releases, so the JSON schema is not entirely correct (it only accepts `version_overrides` for GitHub release sources), but I decided I don't care.


williambotman:

PR #740, opened 2023-03-02
remove haskell-language-server

williamboman:
This is best installed via ghcup, which is not supported.


williambotman:

PR #903, opened 2023-03-19
feat: add clippy

mrtnvgr:


williambotman:

williamboman:
Hello and congrats on the first external contribution! It seems to be struggling with the version specifier:

```
error: the `--version` provided, `v0.0.302`, is not a valid semver version: cannot parse 'v0.0.302' as a semver
```

packages/clippy/package.yaml
```diff
@@ -0,0 +1,17 @@
...
+licenses:
+  - Apache-2.0
+  - MIT
+languages:
+  - Rust
+categories:
+  - Linter
+
+source:
+  id: pkg:cargo/clippy@v0.0.302
```
williamboman, line 14: This ought to fix it
```suggestion
  id: pkg:cargo/clippy@0.0.302
```

mrtnvgr:
> Installing clippy via crates.io is deprecated. Please use the following:
> 
> rustup component add clippy-preview
> on a Rust version 1.29 or later. You may need to run rustup self update if it complains about a missing clippy binary.
> 
> See [the homepage](https://github.com/rust-lang-nursery/rust-clippy/#clippy) for more information

mrtnvgr:
```lua
Caused by:
  process didn't exit successfully: `rustc - --crate-name ___ --print=file-names -Zunstable-options -Zbinary-dep-depinfo --crate-type bin --crate-type rlib --crate-type dylib --crate-type cdylib --crate-type staticlib --crate-type proc-macro --print=sysroot --print=cfg` (exit status: 1)
  --- stderr
  error: the option `Z` is only accepted on the nightly compiler
```

Clippy requires nightly rustc.

mrtnvgr:
It seems to me that the clippy can be installed only via `rustup`... What we can do?

williamboman:
> It seems to me that the clippy can be installed only via `rustup`... What we can do?

Hm I see. Packages that are managed via `rustup` is not supported (there are no plans to support it either). One would have to install and manage it directly via `rustup`

PR #924, opened 2023-03-21
feat: add stylelint

benfrain:
Apologies in advance – first time doing this! Hopefully this is in the correct format.

Closes https://github.com/williamboman/mason.nvim/issues/695.

### DIFF
--- a/packages/stylelint/package.yaml
+++ b/packages/stylelint/package.yaml
@@ -1,0 +1,17 @@
+---
+name: stylelint
+description: A mighty CSS linter that helps you avoid errors and enforce conventions.
+homepage: https://stylelint.io
+licenses:
+  - MIT
+languages:
+  - CSS
+  - Sass
+  - SCSS
+  - LESS
+categories:
+  - Linter
+source:
+  id: pkg:github.com/stylelint/stylelint/releases/tag/15.3.0
+bin:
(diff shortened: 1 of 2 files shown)


williamboman (commented):
Hey! It seems like stylelint should preferably be installed locally in each project as a dev dependency, probably to make it work with its plugin system. There are other packages that have the same "problem", so I'm not opposed to adding stylelint, just thought I'd point that out.

benfrain:
Hmmm, that’s a very good point. I know null-ls has the ‘prefer_local’ option for sources so that should mean that people can still use a different stylelint version for their own project and fallback to the Mason one if none more local? Guess if going that route, a little info in the readme might be useful. I know I in the past I spent more time than I care to remember trying to debug a prettier issue with a version local and another in root before I realised about the ‘prefer_local’ option in null-ls. 
By the same token, for this project, it may be more appropriate to not add it to Mason but add a note as to why it should be installed local only 🤷‍♂️

chrisgrieser:
just to add to this, I just tried the mason version, and while it works nicely, it does not play well with plugins installed via a different method (npm). 

So other either you use no plugins, or you have to install stylelint from somewhere else regardless.

though adding the most popular plugins like stylelint-order might alleviate this issue at least.

PR #1160, opened 2023-04-12
feat: add helm-ls

DrummyFloyd:
i found an early state project, that provide a Language server for Helm file.
even if it's in early dev, it work really good.
would be good to have this LSP in Mason.

especially for DevOps ppl.

### DIFF
--- a/packages/helm-ls/package.yaml
+++ b/packages/helm-ls/package.yaml
@@ -1,6 +1,6 @@
 ---
-name: helm_ls
-description: A language server that offers Helm support in early development- programmed in Go.
+name: helm-ls
+description: A language server that offers Helm support in early development - programmed in Go.
 homepage: https://github.com/mrjosh/helm-ls
 licenses:
   - MIT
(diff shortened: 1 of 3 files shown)


williamboman (approved):
Thanks! I changed the name to `helm-ls` match the upstream name. I also removed the `bin:` entries as they're not needed (they are all the same as the `file:` value).

DrummyFloyd:
thank you for your rework! andfor the acceptance !

DrummyFloyd:
@williamboman  do i need to do somethings else, to have a functionnal `helm-ls`
 like add a rule for starting the server `helm-ls start` in mason.nvim repo ?

williamboman:
Their README include an example: https://github.com/mrjosh/helm-ls#nvim-lspconfig-setup. It might make sense to contribute the server configuration to nvim-lspconfig itself

DrummyFloyd:
dumb question, but if do this contrbution 
mason + mason-lspconfig will work with helm-ls ?

williamboman:
Mason only installs packages on your system. These packages can be used by anyone once installed. Try for example invoking the `helm_ls` executable via Neovim's shell interface: `:!helm_ls --help`. To start a server via Neovim's builtin LSP client you can call `:h vim.lsp.start()`, for example:

```lua
vim.lsp.start {
  name = "helm language server",
  cmd = { "helm_ls", "serve" }, -- helm_ls will need to exist in PATH, which Mason will ensure if you've installed it via Mason.
  root_dir = vim.fs.dirname(vim.fs.find({'Chart.yaml'}, { upward = true })[1]),
}
``` 

nvim-lspconfig is an optional plugin that primarily:
1) provides a bunch of "default" server configurations to help users easily use a server
2) provides some boilerplate code for finding root directory and convenience commands like `:LspStart`, `:LspStop`, `:LspInfo`

nvim-lspconfig will call `vim.lsp.start()` under the hood for you.

williamboman:
And mason-lspconfig.nvim primarily:
1) hosts a mapping between an nvim-lspconfig server name and a Mason package name
2) includes extra, necessary, configuration to make a server installed through Mason compatible with the default server configurations
3) provides utilities to automatically set up servers installed via Mason

PR #1198, opened 2023-04-14
feat: add millet lsp

igorlfs:
Closes williamboman/mason.nvim#1097.

### DIFF
--- a/packages/millet/package.yaml
+++ b/packages/millet/package.yaml
@@ -19,16 +19,16 @@ source:
     - target: win_arm64
       file: millet-ls-aarch64-pc-windows-msvc.gz
       bin: millet-ls-aarch64-pc-windows-msvc
-    - target: linux_arm64
+    - target: linux_arm64_gnu
       file: millet-ls-aarch64-unknown-linux-gnu.gz
       bin: millet-ls-aarch64-unknown-linux-gnu
     - target: darwin_x64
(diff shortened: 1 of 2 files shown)


packages/millet/package.yaml
```diff
@@ -0,0 +1,36 @@
+---
+name: millet
+description: A language server for Standard ML.
+homepage: https://github.com/azdavis/millet
+licenses:
+  - Apache-2.0
+  - MIT
+languages:
+  - SML
```
williamboman, line 9:
```suggestion
  - Standard ML
```

schemas/enums/language.json
```diff
@@ -126,6 +126,7 @@
         "Shell",
         "Slint",
         "Smithy",
+        "SML",
```
williamboman, line 129:
```suggestion
        "Standard ML",
```

williamboman (commented):
I think this abbreviation could be expanded to aid discovery, wdyt?

igorlfs:
> I think this abbreviation could be expanded to aid discovery, wdyt?

Sure!

schemas/enums/language.json
williamboman, line 129:
```suggestion
        "Standard ML",
```

williamboman (approved):
Great!

PR #1217, opened 2023-04-16
feat: add prettier-eslint

jnhooper:
Sorry I couldn't find any etiquette for the PR template but I'm trying to add [prettier-eslint](https://github.com/prettier/prettier-eslint)
I'm not quite sure how to test this locally... i tried adding my fork to the registries list but i get an error that it failed
```
Registry installation failed with the following error:
    GitHubRegistrySource(repo=jnhooper/mason-registry) failed to install: Failed to download registry archive.```

### DIFF
--- a/packages/prettier-eslint/package.yaml
+++ b/packages/prettier-eslint/package.yaml
@@ -1,0 +1,29 @@
+---
+name: prettier-eslint
+description: Prettier followed by eslint --fix
+homepage: http://github.com/prettier/prettier-eslint
+licenses:
+  - MIT
+languages:
+  - Angular
+  - CSS
+  - Flow
+  - GraphQL
+  - HTML
+  - JSON
+  - JSX
+  - JavaScript
+  - LESS
+  - Markdown
+  - SCSS
+  - TypeScript
+  - Vue
+  - YAML
+categories:
+  - Formatter
+
+source:
+  id: pkg:npm/prettier-eslint@15.0.1
+
+bin:
+  prettier-eslint: npm:prettier-eslint


williamboman (commented):
Hey! Per their documentation it should be installed as a dev dependency, I can't find any mentions or references to installing it globally, presumably because of how its plugin system works. Any particular reason you want to manage it via Mason?

jnhooper:
I guess i just maybe misunderstood how to work with mason. If i install this locally i would want mason to have both prettier and eslint installed?

williamboman:
> I guess i just maybe misunderstood how to work with mason. I was updating my workflow to use NvChad and trying to get this to work in that environment. If i install prettier-eslint locally then i would want mason to have both prettier and eslint installed, does that sound right?

I don't know what the convention is with this particular package. Generally you want an environment that is reproducible across environments, which usually means installing these as devDependencies in the project itself. This will install executables to `node_modules/.bin`, but not to PATH (unless explicitly configured). null-ls has the `prefer_local` flag that one can enable on certain sources to prefer project-local installations.

I'd be happy to merge this but I'm a bit worried about ending up promoting incorrect ways of using the package.

jnhooper:
> > I guess i just maybe misunderstood how to work with mason. I was updating my workflow to use NvChad and trying to get this to work in that environment. If i install prettier-eslint locally then i would want mason to have both prettier and eslint installed, does that sound right?
> 
> I don't know what the convention is with this particular package. Generally you want an environment that is reproducible across environments, which usually means installing these as devDependencies in the project itself. This will install executables to `node_modules/.bin`, but not to PATH (unless explicitly configured). null-ls has the `prefer_local` flag that one can enable on certain sources to prefer project-local installations.
> 
> I'd be happy to merge this but I'm a bit worried about ending up promoting incorrect ways of using the package.

nah dont worry about it, i wouldn't want to start a bad habit. I'll close the PR and keep thinking about this a bit

PR #1240, opened 2023-04-17
feat(netcoredbg): Update package.yaml to include linux x64 arch

mebaran:
Added x64 as valid arch for linux (pre-built release)

### DIFF
--- a/packages/netcoredbg/package.yaml
+++ b/packages/netcoredbg/package.yaml
@@ -17,10 +17,10 @@ source:
     - target: [darwin_arm64, darwin_x64]
       file: netcoredbg-osx-amd64.tar.gz:libexec/
       bin: exec:libexec/netcoredbg/netcoredbg
-    - target: linux_arm64
+    - target: linux_arm64_gnu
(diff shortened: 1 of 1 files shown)


packages/netcoredbg/package.yaml
```diff
@@ -20,6 +20,9 @@ source:
     - target: linux_arm64
       file: netcoredbg-linux-arm64.tar.gz:libexec/
       bin: exec:libexec/netcoredbg/netcoredbg
+    - target: linux_x64
```
williamboman, line 23: Uses GNU linker:

```suggestion
    - target: linux_x64_gnu
```

packages/netcoredbg/package.yaml
```diff
@@ -20,6 +20,9 @@ source:
     - target: linux_arm64
```
williamboman, line 20:
```suggestion
    - target: linux_arm64_gnu
```

williamboman (commented):
Thanks! Just one change needed to target GNU linux

PR #1294, opened 2023-04-20
feat: add sonarlint-language-server

huesersohn:


packages/sonarlint-language-server/package.yaml
```diff
@@ -0,0 +1,32 @@
...
+source:
+  id: pkg:github/SonarSource/sonarlint-vscode@3.17.0
+  asset:
+    file: sonarlint-vscode-{{version}}.vsix
+
+bin:
+  sonarlint-language-server: java-jar:extension/server/sonarlint-ls.jar
+
+share:
+  sonarlint-analyzers: extension/analyzers
```
williamboman, line 32:
```suggestion
  sonarlint-analyzers: extension/analyzers/
```

williamboman (approved):
Thanks! I think it might make sense to add support for a `mvn` package type as jars are published on Maven. This asset file contains a bunch of additional, unused, files of ~100 MiB

huesersohn:
The bulk of the size (124 MB) comes from the analyzers that I link to shared as they are needed for the various filetypes. Apart from that the Server is 23 MB, so that leaves about 3 or 4 MB for the rest of the files.

williamboman:
> The bulk of the size (124 MB) comes from the analyzers that I link to shared as they are needed for the various filetypes. Apart from that the Server is 23 MB, so that leaves about 3 or 4 MB for the rest of the files.

Ah, I see 👍

williamboman (approved):
Thanks again, on mobile so had to rely on CI for the needed adjustments. Should pass now 🤞

huesersohn:
Thank you!

PR #1295, opened 2023-04-20
feat: add tfsec

op:
This adds `tfsec` to the mason registry, used by `null-ls`.

I'm unsure how to test this locally so beware. :)


williamboman (approved):
Thanks!

PR #1322, opened 2023-04-23
feat: add solidity-language-server 

anajuliabit:
Others solidity language servers don't work correctly.
See https://github.com/ethereum/solidity/issues/12825


williamboman:
@williambotman /apply

```diff
diff --git a/packages/solidity-language-server/package.yaml b/packages/nomicfoundation-solidity-language-server/package.yaml
similarity index 100%
rename from packages/solidity-language-server/package.yaml
rename to packages/nomicfoundation-solidity-language-server/package.yaml
```

williamboman (approved):
Thanks! I fixed an issue with the purl, where the `@` sign needs to be percent-encoded. I also changed the bin name to `nomicfoundation-solidity-language-server` as it seems to be the established name upstream (I also changed the package name to follow mason's naming guidelines). Let me know if you think these naming changes doesn't make sense!

packages/nomicfoundation-solidity-language-server/package.yaml
```diff
@@ -0,0 +1,16 @@
...
+homepage: https://github.com/NomicFoundation/hardhat-vscode
+licenses:
+  - MIT
+languages:
+  - Solidity
+categories:
+  - LSP
+
+source:
+  id: pkg:npm/%40ignored/solidity-language-server@0.6.14
```
williamboman, line 13: I might be missing something, but the `@ignored` scope doesn't seem to be an official one?

```suggestion
  id: pkg:npm/%40nomicfoundation/solidity-language-server@0.6.16
```

anajuliabit: You are right, changed to oficial

PR #1364, opened 2023-04-26
feat: Add debug-adapter-haskell

sjcobb2022:
Hi there,

I took some time to put together this port of [phoityne/haskell-debug-adapter](https://github.com/phoityne/haskell-debug-adapter).

I would love for any feedback and/or criticisms so that I can improve the.

Like the haskell-language-server package, haskell-debug-adapter is dependent on ghcup. The package uses ghcup to install a local copy of stack. Stack is then used to build local versions of haskell-dap, ghci-dap, and finally haskell-debug-adapter.

I was unable to test my the windows implementation as I do not have a windows machine with me at the moment. I would greatly appreciate anyone who would be able to test this.

### DIFF
--- a/packages/haskell-debug-adapter/package.yaml
+++ b/packages/haskell-debug-adapter/package.yaml
@@ -25,7 +25,7 @@ source:
 
     - target: win
       run: |
-        stack install haskell-dap ghci-dap haskell-debug-adapter-"$env:VERSION" --local-bin-path=($pwd).path
+        stack install haskell-dap ghci-dap haskell-debug-adapter-($env:VERSION) --local-bin-path=($pwd).path
       staged: false
       env:
         VERSION: "{{version}}"
--- a/packages/haskell-debug-adapter/package.yaml
+++ b/packages/haskell-debug-adapter/package.yaml
@@ -15,29 +15,24 @@ source:
   build:
     - target: unix
       run: |
-        ghcup install stack latest -i "$PWD"
-        ./stack install haskell-dap ghci-dap haskell-debug-adapter-"$VERSION" --local-bin-path="$PWD"
+        stack install haskell-dap ghci-dap haskell-debug-adapter-"$VERSION" --local-bin-path="$PWD"
       staged: false
       env:
         VERSION: "{{version}}"
       bin:
-        stack: stack
         ghci-dap: ghci-dap
         haskell-debug-adapter: haskell-debug-adapter
 
     - target: win
       run: |
-        ghcup install stack latest -i ($pwd).path
-        stack.exe install haskell-dap ghci-dap haskell-debug-adapter-"$env:VERSION" --local-bin-path=($pwd).path
+        stack install haskell-dap ghci-dap haskell-debug-adapter-"$env:VERSION" --local-bin-path=($pwd).path
       staged: false
       env:
         VERSION: "{{version}}"
       bin:
-        stack: stack.exe
         ghci-dap: ghci-dap.exe
         haskell-debug-adapter: haskell-debug-adapter.exe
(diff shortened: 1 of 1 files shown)


williamboman (commented):
Thanks! Installing as well as linking Stack seems a bit excessive, no? I think a more suitable trade off would be to require (i.e. assume) `stack` to be installed on the system.

sjcobb2022:
@williamboman Alright changes made!

sjcobb2022:
Just caught this bug as I didn't realized the tests ran earlier. now uses ['haskell-debug-adapter'] instead of haskell-debug-adapter because it was considering it as a mathematical expression.

sjcobb2022:
I'll try and fix the windows issues in a bit.

williamboman:
The necessary CI setup added [here](https://github.com/mason-org/actions/commit/633f9cf157b68ef57b4a16c2dc40e9d581d16d4b). Will rerun CI, hopefully should be fine!

sjcobb2022:
> The necessary CI setup added [here](https://github.com/mason-org/actions/commit/633f9cf157b68ef57b4a16c2dc40e9d581d16d4b). Will rerun CI, hopefully should be fine!

Awesome!

williamboman:
(In the future these external dependencies will be derived and/or explicitly defined in the package definitions themselves, but for now it has to be managed in the action workflow)

sjcobb2022:
> (In the future these external dependencies will be derived and/or explicitly defined in the package definitions themselves, but for now it has to be managed in the action workflow)

sjcobb2022:
Oops sorry mispressed the close with commend button.

There is a bug in the windows build still that I didn't catch.

williamboman:
The syntax was correct, the [build is failing on Windows](https://github.com/mason-org/mason-registry/actions/runs/4822968362/jobs/8591635042) with:

```
Error: [S-8506]

Error parsing targets: Unknown local package: C
```

sjcobb2022:
> The syntax was correct, the [build is failing on Windows](https://github.com/mason-org/mason-registry/actions/runs/4822968362/jobs/8591635042) with:
> 
> ```
> Error: [S-8506]
> 
> Error parsing targets: Unknown local package: C
> ```

Yeh lates push should fix that.

sjcobb2022:
> > The syntax was correct, the [build is failing on Windows](https://github.com/mason-org/mason-registry/actions/runs/4822968362/jobs/8591635042) with:
> > 
> > ```
> > Error: [S-8506]
> > 
> > Error parsing targets: Unknown local package: C
> > ```
> 


 Yeh latest push should fix that.

Sorry with all the hassle i dont have a windows machine to test on so I've been using an online repl to test sorry!

williamboman:
> Yeh lates push should fix that.

Ah interesting, I figured they'd be the same. Do you happen to know what the difference is or what was wrong with `($pwd).path`? My Powershell skills are very basic.

> Sorry with all the hassle i dont have a windows machine to test on so I've been using an online repl to test sorry!

No worries, Windows is always a headache.

williamboman:
Thanks 😊!

sjcobb2022:
> Ah interesting, I figured they'd be the same. Do you happen to know what the difference is or what was wrong with `($pwd).path`? My Powershell skills are very basic.

I think the string interpolation version that I used ensures that the value is a string.

The pwd command on windows is actually an alias to the Get-Location command which has a PathInfo type output.

https://learn.microsoft.com/en-us/powershell/module/microsoft.powershell.management/get-location?view=powershell-7.3#outputs

So I think it wasn't being converted to a string properly.

So I think what was happening it was setting

PR #1388, opened 2023-04-28
bump(staticcheck): update to 2023.1.3 (v0.4.3)

XenoPhex:
See https://github.com/dominikh/go-tools/releases for information about the last several releases

Also worth noting, staticcheck changed the format of it's release names: https://github.com/dominikh/go-tools#releases


williamboman (approved):
They seem to also provide a comprehensive set of prebuilt binaries, might be a good idea to switch to them!

PR #1434, opened 2023-05-02
feat: support rustfmt on more platforms

joske:
On linux ARM64 (rpi or asahi linux or even in docker inside an M1 mac), rustfmt can not be installed. This is because `rustfmt` only releases binaries for linux/windows. No idea how this can ever work on macOS as the github repo doesn't have these artefacts.

I'm trying to fix this by building from source on all platforms. I don't know if this works as I can not really test it. Please let me know if/how I can do it better.

cheers,
jos

### DIFF
--- a/packages/rustfmt/package.yaml
+++ b/packages/rustfmt/package.yaml
@@ -9,19 +9,17 @@ languages:
   - Rust
 categories:
   - Formatter
-
 source:
-  id: pkg:github/rust-lang/rustfmt@v1.5.1
-  asset:
-    - target: [darwin_x64, darwin_arm64]
-      file: rustfmt_macos-x86_64_{{version}}.tar.gz
-      bin: rustfmt_macos-x86_64_{{version}}/rustfmt
-    - target: linux_x64
-      file: rustfmt_linux-x86_64_{{version}}.tar.gz
-      bin: rustfmt_linux-x86_64_{{version}}/rustfmt
-    - target: win_x64
-      file: rustfmt_windows-x86_64-msvc_{{version}}.zip
-      bin: rustfmt_windows-x86_64-msvc_{{version}}/rustfmt.exe
-
+  # renovate:datasource=github-tags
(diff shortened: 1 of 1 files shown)


williamboman (commented):
Hey! Thanks for the PR! Generally, building from source is avoided unless there are no other viable options, and even then package requests that need to built from source may be denied if the build process is too complex (which I should point it's not in this PR). This is a trade-off that is made according to the guiding principles that:

1) Available packages should be available for the general use case (i.e. cover a majority of systems<sup>1</sup>). It's currently acceptable if a package is not supported on a certain platform (e.g. arm systems).
2) Package installation should be reasonably fast (seconds rather than minutes).

This is certainly far from perfect but until there's better support for building from source it's the principle that is followed. Some examples of improved build support are:

- some kind of build harness that for example can ensure build dependencies are available
- providing prebuilt package artifacts via edge cached CDN

For now, the best way to get arm support for this package would be to see if they'd be open to providing prebuilt arm binaries upstream. Otherwise people will have to manage this package outside of mason.

<sup>1 - although there's no data available on what systems people use, I'd wager that Linux x64, GNU Linux x64, and Darwin (macOS) covers >50% of users at the moment</sup>

joske:
Hi @williamboman, I just got feedback from `rustfmt` team that they won't be providing any binaries for any platform anymore going forward (see the above PR). 

This means ALL platforms will not get any updates anymore. I'd guess this is not acceptable.

Do you have a suggestion on how to make progress on this?

joske:
IIUC, the binaries would depend on system rust libraries, so they wouldn't work anyway without having rust installed. 

But, as the maintainer says, it's unlikely that anyone would want/need `rustfmt` without having rust installed. 

Can we copy rustfmt from system $PATH if available?

joske:
BTW, building from source took 19 seconds on my system (M2 MBP).

joske:
In fact, `rustfmt` is already broken. The 1.6.0 release is already linked to librustc_driver. My code is no longer being formatted. I'll log an issue.

PR #1440, opened 2023-05-02
feat: add texlab on arm64 linux

joske:
On ARM64 linux, texlab is not installable, while packages exist in github. This PR adds support for linux ARM64.


williamboman (approved):
Thanks! I noticed these are linked to GNU linker, so I added _gnu specifiers to the Linux targets

PR #1465, opened 2023-05-04
Update netcoredbg to respect windows style paths

ghost-pep:
## Overview
Before, the path to the executable for netcoredbg was using `/` rather than `\`. This led to DAP never launching netcoredbg and not working for windows users. This PR fixes that!

## How I got here
I have been using [LazyVim's DAP setup](https://www.lazyvim.org/plugins/extras/dap.core) for out of the box DAP configuration. Launching netcoredbg results in a failure to connect and I found that the path to the executable is being launched like so:
image
Thus, I have created this fix PR.

### DIFF
--- a/packages/netcoredbg/package.yaml
+++ b/packages/netcoredbg/package.yaml
@@ -25,7 +25,7 @@ source:
       bin: exec:libexec/netcoredbg/netcoredbg
     - target: win_x64
       file: netcoredbg-win64.zip
-      bin: netcoredbg/netcoredbg.exe
+      bin: netcoredbg\netcoredbg.exe
 
 bin:
   netcoredbg: "{{source.asset.bin}}"


ghost-pep:
Hmmm this actually needs more debugging on my end. Seems like the path is fine mixing the two styles. I am going to close this for now as I now think the problem is with DAP configuration and not acquisition of the netcoredbg binary itself 🤷

williamboman:
Hello! Paths in package definitions use the `/` path separator everywhere. The idea is to normalize these when paths are constructed in the client (mason.nvim). However my understanding is that mixing `\` and `/` should be perfectly fine for most use cases on Windows. I also believe that these paths are "expanded" before they could pose any problems with underlying tools, but I may be wrong here.

PR #1469, opened 2023-05-04
elixir-ls v0.14.5

Miradorn:
there is no actual release but only tags for the last two releases, so renovate hasn't picked them up i think?


williamboman:
If there's no release then that version is incompatible. I saw they drafted releases which just got picked up by renovate. Thanks!

PR #1503, opened 2023-05-07
Pull request to add Python docstring formatters

kylefhartzenberg:
Pull request to add two of the major Python docstring formatting packages to the mason registry, namely:
- [docformatter](https://pypi.org/project/docformatter/), and
- [pyment](https://pypi.org/project/pyment/).

Both packages provide functionality to format docstrings in Python files, with each package providing relative advantages/disadvantages over the other. I expect development pace to vary between the two and that mason users require the option to select a formatter that suits their project's needs. The closest existing package in the [mason registry](https://mason-registry.dev/registry/list) is pydocstyle, however, that package only provides PEP 257 compliance checking, not formatting.

The added package.yaml files follow the package definition scheme, and have been created in unique directories ("docformatter" and "pyment") in accordance with the [contribution guidelines](https://github.com/mason-org/mason-registry/blob/main/CONTRIBUTING.md).

### DIFF
--- a/packages/docformatter/package.yaml
+++ b/packages/docformatter/package.yaml
@@ -1,0 +1,16 @@
+---
+name: docformatter
+description: docformatter automatically formats docstrings to follow a subset of the PEP 257 conventions.
+homepage: https://pypi.org/project/docformatter/
+licenses:
+  - MIT
+languages:
+  - Python
+categories:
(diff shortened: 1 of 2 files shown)


williamboman (approved):
Awesome, thanks!

PR #1576, opened 2023-05-13
feat: add prisma

Pandoks:
The formatting for prisma is within the prisma package itself `prisma format <file>`.

### DIFF
--- a/packages/prisma-fmt/package.yaml
+++ b/packages/prisma-fmt/package.yaml
@@ -1,0 +1,17 @@
+---
+name: prisma-fmt
+description: |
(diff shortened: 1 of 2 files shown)


williamboman (commented):
Hey so the prisma CLI is generally installed locally in a project (which is also recommended in their docs). It's also a pretty universal executable in the sense that it does a lot of things other than just providing a formatting facility. I'm a bit hesitant to adding it to the registry for these reasons.

I'm still trying to find out where the `prisma-fmt` executable comes from, which null-ls uses in their formatting source. Do you know?

PR #1592, opened 2023-05-15
add nimlangserver

daylinmorgan:
Adds support for [`nim-langserver`](https://github.com/nim-lang/langserver).

### DIFF
--- a/packages/nimlangserver/package.yaml
+++ b/packages/nimlangserver/package.yaml
@@ -1,5 +1,5 @@
 ---
-name: nim-langserver
+name: nimlangserver
 description: The Nim language server implementation (based on nimsuggest)
 homepage: https://github.com/nim-lang/langserver
 licenses:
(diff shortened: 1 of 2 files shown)


williamboman (commented):
Thanks! Do you think they'd be open to providing immutable release versions, preferably following semver? The usage of the `latest` tag is a bit problematic due to how it's a mutable tag that they delete and retag.

packages/nimlangserver/package.yaml
```diff
@@ -0,0 +1,24 @@
...
+licenses:
+  - MIT
+languages:
+  - Nim
+categories:
+  - LSP
+
+source:
+  # renovate:datasource=github-tags
+  id: pkg:github/nim-lang/langserver@latest
```
williamboman, line 14: I see you've opened an issue with them regarding stable tags. For now let's target commit SHAs so we can get this merged

```suggestion
  # renovate:datasource=git-refs
  id: pkg:github/nim-lang/langserver@f7b32fa651622d77f2da7b357cb152c58fe7b359
```

PR #1609, opened 2023-05-17
fix smithy language server version

louishuyng:
## Problem

I attempted to build the current Smithy language server using the two most recent Java versions, 17 and 20, but neither was successful.

### Old version (e4675485051c4931381470299a7d3ce50408f7c7)
image
image

### New version (708bfc4ff8ca1b266a66c9919d269cd74dffe2e1) based on the latest comment
image
image

### DIFF
--- a/packages/smithy-language-server/package.yaml
+++ b/packages/smithy-language-server/package.yaml
@@ -11,7 +11,7 @@ categories:
 
 source:
   # renovate:datasource=git-refs
-  id: pkg:github/awslabs/smithy-language-server@e4675485051c4931381470299a7d3ce50408f7c7
+  id: pkg:github/awslabs/smithy-language-server@708bfc4ff8ca1b266a66c9919d269cd74dffe2e1
   build:
     run: ./gradlew --no-daemon build
 


williamboman (approved):
Thanks! I wonder why Renovate didn't pick up these new digests. It would be nice if they were to actually tag stable releases, if you feel like raising an issue with them (them being smithy-language-server).

williamboman:
So whenever there's a change to a package definition (anything under `packages/`) there will be a new [release ](https://github.com/mason-org/mason-registry/releases) made immediately. After merging a PR it takes a couple of seconds to actually publish the release, and then a couple more seconds (or minutes) for APIs involved to revalidate their cache.

The registry used by `mason.nvim` is basically updated automatically 0-2 times per day (depends on how much you interact with `mason.nvim`). You can always manually update the registry by doing `:MasonUpdate` (or pressing `C` to check for new package versions when the `:Mason` window is open).

louishuyng:
Awesome Thanks so much for your insight explanation 🙏

PR #1641, opened 2023-05-21
feat: add pest-language-server

Samyak2:
Closes williamboman/mason.nvim#1313


Samyak2:
This LSP only works on `.pest` files, but the validation is failing because there's no language named "Pest". How do I fix this?

packages/pest-language-server/package.yaml
```diff
@@ -0,0 +1,16 @@
+---
+name: pest-language-server
+description: A language server for Pest grammar.
+homepage: https://github.com/pest-parser/pest-ide-tools/tree/main
```
williamboman, line 4:
```suggestion
homepage: https://github.com/pest-parser/pest-ide-tools
```

williamboman:
> This LSP only works on `.pest` files, but the validation is failing because there's no language named "Pest". How do I fix this?

It needs to be included in the enum specification in the schema, I pushed a commit that adds this!

PR #1653, opened 2023-05-22
feat: add lean-language-server

MithicSpirit:
Closes https://github.com/williamboman/mason.nvim/issues/1317

### DIFF
--- a/packages/lean-language-server/package.yaml
+++ b/packages/lean-language-server/package.yaml
@@ -1,0 +1,16 @@
+---
+name: lean-language-server
+description: Lean3 Language Server.
+homepage: https://www.npmjs.com/package/lean-language-server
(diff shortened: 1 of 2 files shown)


williamboman (approved):
Thanks! I changed the language to "Lean" and added it to the `language.json` schema enum so that it's valid. The rationale behind changing to Lean is that these language entries are just meant as a high level element that aids discovery of packages - am I right in assuming there's no significant difference between "Lean 3" and "Lean"?

MithicSpirit:
Oops, didn't realize that I had to add to the list of languages. Also there is a significant difference between Lean3 and Lean4, so I think that it might be good to keep them separate. `lean-language-server` does not work for Lean4.

PR #1780, opened 2023-06-03
feat: add markmap-cli

jay-babu:


williamboman (approved):
Thanks! I fixed the file name extension and removed Runtime from categories (this category is meant for runtimes like Node, Deno, Python)

jay-babu:
Thanks! I will fix the failure in a bit. I thought >0 category was required, but none really fit

PR #1790, opened 2023-06-05
adding ts-standard

kumkee:
Adding ts formatter, https://github.com/standard/ts-standard.


packages/standardts/package.yaml
```diff
@@ -0,0 +1,17 @@
+---
+name: ts-standard
+description: Typescript style guide, linter, and formatter using StandardJS.
+homepage: https://github.com/standard/ts-standard
+licenses:
+  - MIT
+languages:
+  - Typescript
```
williamboman, line 8:
```suggestion
  - TypeScript
```

PR #1809, opened 2023-06-07
Add rubyfmt

hahuang65:


williamboman (approved):
Thanks! I changed rubyfmt to pull release assets instead of building from source. Building from source requires rust and make to be installed, I believe using prebuilt assets will enable more people to install it.

PR #1817, opened 2023-06-08
bump(helm-ls): downgrade to 0.0.3

alesbrelih:
It was using invalid helm-ls version. Version 0.0.4 does not exist.


alesbrelih:
The least I can do...

Thank you for all the effort put into this!

PR #1880, opened 2023-06-15
chore: update vue-language-server

kalvenschraut:
Dependency bot doesn't seem to be pulling in the "pre-release" changes in, for volar, but was manually moved to point to 1.7 due to  package name change. Need to update to latest to fix issues with typescript 5.1.3

### DIFF
--- a/packages/vue-language-server/package.yaml
+++ b/packages/vue-language-server/package.yaml
@@ -10,7 +10,7 @@ categories:
   - LSP
 
 source:
-  id: pkg:npm/%40vue/language-server@1.7.13
+  id: pkg:npm/%40vue/language-server@1.7.0
   extra_packages:
     - typescript
 
--- a/packages/vue-language-server/package.yaml
+++ b/packages/vue-language-server/package.yaml
@@ -10,7 +10,7 @@ categories:
   - LSP
 
 source:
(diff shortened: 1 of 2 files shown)


williamboman (commented):
Hey, thanks! It seems like they might be tagging releases incorrectly. The `latest` dist-tag in npm is `1.7.0` (`next` is  `1.7.13`). `1.7.13` looks to me like a stable release and should be `latest`. Might be worth filing an issue upstream instead?

johnsoncodehk:
v1.7 is indeed a pre-release version, it releases "next" with an odd minor version number in order to accommodate VSCode Markerplace that does not support semantic version tags.

williamboman:
> v1.7 is indeed a pre-release version, it releases "next" with an odd minor version number in order to accommodate VSCode Markerplace that does not support semantic version tags.

Ah I see. I feel like the npm package should follow a different versioning scheme because it's after all a generic distribution for the LSP server.

Anyways, I think we should change Renovate to pick up "unstable" versions for this package instead, to ensure future versions are picked up. I pushed a commit for this, hopefully it works first try (I haven't yet figured out a way to test renovate locally so we'll have to do it live)

williamboman:
Ok after a couple of tries there we go: #1886

PR #2010, opened 2023-06-29
feat: add vale-ls

williamboman:
Closes https://github.com/williamboman/mason.nvim/issues/1383.


packages/vale-ls/package.yaml
```diff
@@ -0,0 +1,34 @@
+---
+name: vale-ls
+description: An implementation of the Language Server Protocol (LSP) for the Vale command-line tool.
+homepage: https://vale.sh/
+licenses:
+  - MIT
+languages:
+  - Text
+  - Markdown
+  - LaTeX
```
williamboman, line 10:
```suggestion
```

PR #2013, opened 2023-06-29
feat: add nextls

jared-mackey:
Hi!

First time making one of these. Please let me know of any adjustments needed. I have not been able to test it yet since I am not sure how. Will be working through that while this PR is open.

### DIFF
--- a/packages/elixir-next-ls/package.yaml
+++ b/packages/elixir-next-ls/package.yaml
@@ -1,0 +1,21 @@
+---
+name: elixir-next-ls
+description: |
+  NextLS is the language server for Elixir that just works.
+homepage: https://www.elixir-tools.dev/next-ls/
+licenses:
+  - MIT
+languages:
+  - Elixir
+categories:
+  - LSP
+
+source:
+  id: pkg:github/elixir-tools/next-ls@0.4.0
+  asset:
+    - target: unix
+      file: nextls
+      bin: nextls
+
+bin:
+  nextls: "{{source.asset.bin}}"
--- a/packages/elixir-next-ls/package.yaml
+++ b/packages/elixir-next-ls/package.yaml
@@ -1,7 +1,6 @@
 ---
 name: elixir-next-ls
-description: |
-  NextLS is the language server for Elixir that just works.
+description: NextLS is the language server for Elixir that just works.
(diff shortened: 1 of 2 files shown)


williamboman (commented):
The package definition looks good to me! However, the file that gets downloaded is just a small Elixir script that in turn calls into Mix to install the nextls language server (presumably globally, i.e. outside of Mason), and then starts it. This makes it not a particularly good fit for Mason because it won't really be installing anything via Mason, or rather it will install an install script, which is kinda weird.

mhanberg:
Hi!

NextLS author here. 

I completely agree with what you've said here, and have tried to explain that mason is not as useful here, but...

eventually (in the coming months maybe) I hope to distribute NextLS as a compiler binary, rather than a bootstrapping script. 

So, Mason would be useful then, and I think the mason registry here wouldn't need to change. 

I plan on still having my plugins (neovim, vscode, etc) downloading it for you, but some people prefer mason. 

I hope that helps in your calculus on whether to include NextLS in the registry.

williamboman:
@mhanberg Cool I had some issues with the Burrito wrapper on Ubuntu 20.04, it pulled binaries linked to glibc 2.34 (quite recent release) so I couldn't execute it. Not sure if it's a Burrito issue or nextls issue though.

mhanberg:
That's a burrito thing. 

I'm not sure if there is a way to lower it but I'll talk to the burrito author. 

He was working on embedding musl statically instead of dynamic linking so that might make it easier to support older operating systems.

mhanberg:
@williamboman is that a blocker? Just for clarity, if I need to get it to work on older operating systems.

williamboman:
Hey sorry I got distracted while reviewing this, forgot about this.

My final thought was about the name, I'm leaning towards `nextls` (or `next_ls` or `next-ls`), wdyt? It might cause people to think it's related to Next.js but I think the description and language mapping makes up for that. Also, what is the canonical binary name, `nextls` or `next_ls`?

mhanberg:
@williamboman no worries at all.

let's go with `nextls` for name and binary

mhanberg:
Thanks!

@jared-mackey  let me know if you get a moment to try this out

PR #2019, opened 2023-06-30
Add metals (scala lsp)

AbdelrahmanDwedar:
I'm not very sure if that's all it needs or not, I'm new to this, if there's anything left that needs to be added please inform me.

### DIFF
--- a/packages/metals/package.yaml
+++ b/packages/metals/package.yaml
@@ -1,0 +1,19 @@
+---
+name: metals
+description: Scala language server
+homepage: https://github.com/scalameta/metals
+licenses:
+  - Apache-2.0
+languages:
+  - Scala
+categories:
+  - LSP
+
+source:
+  # renovate:datasource=github-tags
+  id: pkg:github.com/scalameta/metals@0.11.12
+
+schemas:
+  lsp: vscode:https://raw.githubusercontent.com/scalameta/metals-vscode/{{version}}/packages/metals-vscode/package.json
+bin:
+  metals: "{{source.build.bin.lsp}}"


williamboman (commented):
Hello! So this package definition is incomplete. Installing metals is done via concourse, which makes it a bit difficult to add it to mason for reasons stated in this issue https://github.com/williamboman/mason.nvim/issues/369

AbdelrahmanDwedar:
Oh I see. What about [scalafmt](https://scalameta.org/scalafmt/) or [scalastyle](http://www.scalastyle.org/)?

williamboman:
> Oh I see. What about [scalafmt](https://scalameta.org/scalafmt/) or [scalastyle](http://www.scalastyle.org/)?

Scalastyle seems to provide standalone jars at https://repo1.maven.org/maven2/org/scalastyle/scalastyle_2.12/. Scalafmt seems to be the same as metals (also, it seems like metals uses scalafmt for LSP-based formatting, so it might be slightly redundant)

williamboman:
For downloading files from "proprietary" hosts (such as the scalafmt jars location), the `generic` pkg type should be used. See for example how [jdtls](https://github.com/mason-org/mason-registry/blob/main/packages/jdtls/package.yaml) does it

AbdelrahmanDwedar:
Ok, thanks, may I make another PR for adding scalastyle or shall I remove the the metals and add it in the PR?

williamboman:
I think it'd be a good idea to create a new branch and PR for that, makes it easier for future travelers to search and find information regarding metals.

Issue #2054, opened 2023-07-04
`rustfmt` is broken


joske:
Hi,

due to upstream changes, the binaries for `rustfmt` are now broken (they no longer work as standalone binary). See https://github.com/mason-org/mason-registry/pull/1434 for more information.

When I change the symlink in `~/.local/share/nvim/mason/bin/rustfmt` to point to my system `rustfmt`, code formatting works again.

Downloading the binaries from `rustfmt` project is not a viable option going forward.

williamboman:
Hello! Yeah I opened an issue with them to learn more, it seems like `rustfmt` will need to be managed via `rustup` directly onwards due to changes upstream in rustc.

What this means right now is that the `rustfmt` package will have to be deprecated in Mason, and users must manage it externally (i.e. directly via `rustup component`). `rustup component` installs components in `~/.rustup` in the currently active toolchain. Changing Mason to refer to these would be out of scope and violate the principle of fully contained installations - the Mason installation would become "corrupt" should a user change toolchains or uninstall `rustfmt`.

williamboman:
Package has been deprecated and downgraded to a functional version in #2059. Should the distribution alternatives change in the future, it will be "un-deprecated" again 😄

PR #2055, opened 2023-07-04
use system rustup to get rustfmt

joske:
A first stab at creating a new package.yaml for rustfmt as the current one is broken (see https://github.com/mason-org/mason-registry/issues/2054).

I could not figure out how I can test this.

Please comment on how it could be improved.

### DIFF
--- a/packages/rustfmt/package.yaml
+++ b/packages/rustfmt/package.yaml
@@ -11,17 +11,12 @@ categories:
   - Formatter
 
 source:
-  id: pkg:github/rust-lang/rustfmt@v1.6.0
-  asset:
-    - target: [darwin_x64, darwin_arm64]
-      file: rustfmt_macos-x86_64_{{version}}.tar.gz
-      bin: rustfmt_macos-x86_64_{{version}}/rustfmt
-    - target: linux_x64
-      file: rustfmt_linux-x86_64_{{version}}.tar.gz
-      bin: rustfmt_linux-x86_64_{{version}}/rustfmt
-    - target: win_x64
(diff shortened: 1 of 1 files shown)


packages/rustfmt/package.yaml
```diff
@@ -11,17 +11,10 @@ categories:
...
-    - target: linux_x64
-      file: rustfmt_linux-x86_64_{{version}}.tar.gz
-      bin: rustfmt_linux-x86_64_{{version}}/rustfmt
-    - target: win_x64
-      file: rustfmt_windows-x86_64-msvc_{{version}}.zip
-      bin: rustfmt_windows-x86_64-msvc_{{version}}/rustfmt.exe
+  id: pkg:generic/rustfmt@x
+  build:
+    run: cp $(which rustfmt) . || echo "rustfmt not found"
+    bin: rustfmt
```
williamboman, line 17: This is problematic for a couple of reasons:

1) It has a placeholder version number. Packages must refer to real version numbers.
2) The build script simply copies an external installation of `rustfmt` into the Mason installation directory. This is not a self-contained installation, and may also be problematic for loading linked libraries (relative `rpath`).
3) Doesn't support Windows (although this is an easy fix by providing a powershell script as well).

williamboman:
Superseded by #2059

joske:
1. We could easily put the version number of the `rustfmt` release and we could track that
2. I tried this locally and just copying the rustfmt binary into mason works (macOS)
3. I don't know anything about powershell unfortunately

PR #2061, opened 2023-07-04
chore: improve deprecation copy

williamboman:
I should probably proofread these in the future.

### DIFF
--- a/packages/rustfmt/package.yaml
+++ b/packages/rustfmt/package.yaml
@@ -12,7 +12,7 @@ categories:
 
 deprecation:
   since: v1.5.1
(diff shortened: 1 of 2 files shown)


packages/rustfmt/package.yaml
```diff
@@ -12,7 +12,7 @@ categories:
 
 deprecation:
   since: v1.5.1
-  message: rustfmt must now be managed externally via rustup.
+  message: rustfmt should now be installed externally via rustup.
```
williamboman, line 15:
```suggestion
  message: rustfmt should now be installed via rustup.
```

Issue #2083, opened 2023-07-06
Support rubocop as an LSP 


gongfarmer:
Rubocop has long existed in the mason package list, as a linter.

Recent rubocop releases starting at v1.53 now [include built-in lsp support](https://github.com/rubocop/rubocop/pull/11926).

Please support rubocop as an lsp.

PR #2084, opened 2023-07-06
fix(rubocop): rubocop has become an lsp since v1.53 (#2083)

gongfarmer:


williamboman (commented):
Thanks! Is it still not primarily a linter & formatter, and merely provides LSP as a new way of interfacing with it? We can keep the old categories

gongfarmer:
You might be right about that.  I don't know the ramifications of this choice.
I deleted the old categories after looking at the package.yml for solargraph and ruby-lsp, which are declared as LSP only.

How can I test this?  I did not find a copy of any package.yaml in ~/.local/share/nvim or ~/.local/state/nvim.

williamboman:
> You might be right about that. I don't know the ramifications of this choice. I deleted the old categories after looking at the package.yml for solargraph and ruby-lsp, which are declared as LSP only.

These categories are purely for presentational purposes only, so whatever is chosen won't have any "programmatic" impact (unless there's 3rd party or custom code that checks package categories somewhere, but that is a frail approach and should not be done).

Finally, it would be good to provide a mapping in `mason-lspconfig.nvim` so that it understands that the `rubocop` server configuration in nvim-lspconfig can be installed via the `rubocop` Mason package. Here's an example PR: https://github.com/williamboman/mason-lspconfig.nvim/pull/246. Adding the mapping entry to `lua/mason-lspconfig/mappings/server.lua` and then running `make generate` is enough.

williamboman (approved):
Thanks!

PR #2129, opened 2023-07-12
bump(sonarlint-language-server): update to v3.19.2

thahnen:
Hello everyone,

updated to the newest version of SonarLint for Visual Studio Code. Also updated the general package information for the Mason Registry.
Information on provided languages can be found from the *SonarLint Language Server* source code right [HERE](https://github.com/SonarSource/sonarlint-language-server/blob/2.19.1.73109/src/main/java/org/sonarsource/sonarlint/ls/EnginesFactory.java#L50-L69).

Best,
Tobias from SonarLint


williamboman (approved):
Thanks! Btw Renovate is struggling with version upgrades because of the suffix in the tag name. If they were semver-compliant, updates would automatically trickle through.

PR #2138, opened 2023-07-12
feat: add pyright-python

acid-bong:
Reasoning: https://github.com/RobertCraigie/pyright-python is an unofficial Python wrapper for Pyright, that installs node and npm in a small venv in and runs Pyright from there.

It's been discussed earlier in [this issue](https://github.com/williamboman/mason.nvim/issues/424), when the registry was a part of Mason proper. I don't know for how long, but the wrapper has 2 extra executables, `pyright-python` and `pyright-python-langserver`, which are identical to the originally called ones. To avoid conflict with original Pyright, this commit of mine will only expose `..-python` executables to `mason/bin`.

A parallel PR was created at [nvim-lspconfig](https://github.com/neovim/nvim-lspconfig/pull/2718)

### DIFF
--- a/packages/pyright-python/package.yaml
+++ b/packages/pyright-python/package.yaml
@@ -1,0 +1,20 @@
+---
+name: pyright-python
+description: Static type checker for Python - in a Python wrapper.
+homepage: https://github.com/RobertCraigie/pyright-python
+licenses:
+  - MIT
+languages:
+  - Python
+categories:
+  - LSP
+
+source:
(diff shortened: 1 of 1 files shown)


williamboman (commented):
Thanks! The package definition looks good but however this package won't be considered for the registry. It's a very thin CLI wrapper that installs Node & npm for you before installing the actual `pyright` npm package. It's also an unofficial source.

PR #2183, opened 2023-07-17
feat: add pkgbuild-language-server

mehalter:


packages/pkgbuild-language-server/package.yaml
```diff
@@ -0,0 +1,16 @@
+---
+name: pkgbuild-language-server
+description: Language server for ArchLinux/Windows Msys2's PKGBUILD and *.install.
+homepage: https://pkgbuild-language-server.readthedocs.io/en/latest/
+licenses:
+  - GPL-3.0
+languages:
+  - Bash
```
williamboman, line 8: I wonder if PKGBUILD would be a more suitable entry here? It doesn't exist in the schema yet but can be added to `language.json` under `schemas/`

williamboman (commented):
Thanks!

packages/pkgbuild-language-server/package.yaml
mehalter: I think we should still have `Bash` as an entry there but I can also add `PKGBUILD`. `PKGBUILD` is actually a subtype of bash so you can also use the bash language server along side it. What do you think? Here is a snippet from the pkgbuild-language-server README:

> PKGBUILD is a subtype of bash. See [bash-language-server](https://github.com/bash-lsp/bash-language-server) to get support of bash language server.

williamboman: It's not too important that these are taxonomically complete, it's only used to aid searching and discovering packages so either/both works imo!

PR #2191, opened 2023-07-17
feat: add checkstyle

micku:


williamboman (commented):
We probably want to provide some means to access this, probably via an executable:

```yaml
bin:
  checkstyle: java-jar:<path_to_jar>
```

williamboman (approved):
Thanks!

PR #2297, opened 2023-07-27
feat(markdown): add doctoc

pbnj:
CLI for generating markdown table of contents.

Supports github, gitlab, & bitbucket


williamboman (approved):
Thanks and apologies for late response!

PR #2336, opened 2023-08-01
feat: add slang

DoeringChristian:
I have added support for the [Slang](https://github.com/shader-slang/slang) shader language (a derivative of HLSL) to the registry.
I tried to comply with the contribution guidelines, however I'm not sure about the name "slang". It could be seen as a little ambiguous. Since the GitHub organization is named "shader-slang" would that be a better name?

I hope I didn't miss anything, please tell me otherwise.


DoeringChristian:
Btw. I only could test it on Linux for the time being.

williamboman (approved):
This package seems to provide two binaries `slangd` & `slangc`. I think it'd make sense to link them both. I also added a few more targets and fixed the version specifier to target their latest release (not sure where you got your `v0.27.18` version from?).

What do you think?

DoeringChristian:
> 

Yea that makes sense, I focused on the lsp and totally forgot that the compiler is in the same repo.
I also must have overlooked that I was still using an older version.
Thanks for fixing that.

PR #2353, opened 2023-08-03
bump(sonarlint-language-server): update to v3.20.2

thahnen:
Still not semver-compliant 😮 I'm not sure but does the Mend Renovate bot not offer some configuration regarding automatically detecting changes just for one artifact? I'm totally fine with updating myself manually every time, but automation is key nevertheless 😄

### DIFF
--- a/packages/sonarlint-language-server/package.yaml
+++ b/packages/sonarlint-language-server/package.yaml
@@ -27,9 +27,9 @@ categories:
   - Linter
 
 source:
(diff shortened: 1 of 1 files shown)


williamboman (approved):
Thanks! It's the `+74272` suffix that is causing issues with Renovate. By default it'll apply the semver versioning strategy, which will rightfully fail to parse it as a valid semver version. They do provide other versioning strategies (for example `loose`) but I'm not sure any of them would fix this.

PR #2365, opened 2023-08-05
feat: add pico8-ls

kikito:
Adding support for PICO-8 by using https://github.com/japhib/pico8-ls/

Solves https://github.com/japhib/pico8-ls/issues/34
Related - git tags needed in pico8-ls for better renovate support https://github.com/japhib/pico8-ls/issues/45

### DIFF
--- a/packages/pico8-ls/package.yaml
+++ b/packages/pico8-ls/package.yaml
@@ -16,4 +16,4 @@ source:
     run: |
       npm run compile
 bin:
-  pico8_ls: node:client/out-min/main.js
+  pico8-ls: node:client/out-min/main.js
--- a/packages/pico8-ls/package.yaml
+++ b/packages/pico8-ls/package.yaml
@@ -6,7 +6,6 @@ licenses:
   - MIT
 languages:
   - PICO-8
-  - PICO-8 Lua
 categories:
   - LSP
 source:
--- a/packages/pico8-ls/package.yaml
+++ b/packages/pico8-ls/package.yaml
@@ -1,0 +1,20 @@
+---
+name: pico8-ls
+description: PICO-8 Language Server
+homepage: https://github.com/japhib/pico8-ls
+licenses:
+  - MIT
+languages:
+  - PICO-8
+  - PICO-8 Lua
+categories:
+  - LSP
+source:
+  # renovate:datasource=git-refs
+  id: pkg:github/japhib/pico8-ls
+  build:
+    run: |
+      npm run compile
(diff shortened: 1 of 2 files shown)


packages/pico8-ls/package.yaml
```diff
@@ -0,0 +1,20 @@
+---
+name: pico8-ls
+description: PICO-8 Language Server
```
williamboman, line 3: I thought the following description was bit more descriptive 🙈:
```suggestion
description: Full language support for the PICO-8 dialect of Lua.
```

packages/pico8-ls/package.yaml
```diff
@@ -0,0 +1,20 @@
...
+licenses:
+  - MIT
+languages:
+  - PICO-8
+  - PICO-8 Lua
+categories:
+  - LSP
+source:
+  # renovate:datasource=git-refs
+  id: pkg:github/japhib/pico8-ls
```
williamboman, line 14: A version specifier is always necessary, let's use latest `HEAD`:

```suggestion
  id: pkg:github/japhib/pico8-ls@ef01bb52cba6fe358dcc2478ace74033744fdd26
```

packages/pico8-ls/package.yaml
```diff
@@ -0,0 +1,20 @@
...
+  build:
+    run: |
+      npm run compile
+bin:
+  pico8_ls: node:client/out-min/main.js
+
```
williamboman, line 20:
```suggestion
```

williamboman (approved):
I think it'd be much better if they simply provided an npm package for it!

packages/pico8-ls/package.yaml
```diff
@@ -0,0 +1,19 @@
...
+categories:
+  - LSP
+source:
+  # renovate:datasource=git-refs
+  id: pkg:github/japhib/pico8-ls@ef01bb52cba6fe358dcc2478ace74033744fdd26
+  build:
+    run: |
+      npm run compile
+bin:
+  pico8_ls: node:client/out-min/main.js
```
williamboman, line 19: This will be bothersome to change in the future. Seeing as they don't provide an executable yet (so we don't know what name it would potentially be), I think it'd be better to use the more conventional `-` separator.

```suggestion
  pico8-ls: node:client/out-min/main.js
```

kikito:
> I think it'd be much better if they simply provided an npm package for it!

Agreed! Thank you for the merge

PR #2366, opened 2023-08-05
bump(ocamlformat): update to v0.26.0

bdferul:


williamboman (approved):
Thanks! Unfortunately Renovate doesn't support Opam yet.

PR #2373, opened 2023-08-07
Updated ZLS to build directly from source.

Ardelean-Calin:
[ZLS](https://github.com/zigtools/zls) releases seem to have remained stuck to the 0.10.0 version released in **November 2022**.
I have corrected the package manifest so that ZLS is built from source according to the README.md.

**A few questions remain:**
- Is this the appropriate way to define/freeze a ZLS version? By specifying a commit hash (which was the latest commit at the time of this PR.)? If so, would we need to update the version every once-in-a-while?
- What would be the best way to handle the dependency on the latest Zig compiler? Usually the latest Zig Language Server needs the latest Zig compiler to build (0.11.0 as of 7 august 2023). It works on my machine™ since I have the latest Zig compiler, but other people might not have the latest one.

### DIFF
--- a/packages/zls/package.yaml
+++ b/packages/zls/package.yaml
@@ -10,38 +10,15 @@ categories:
   - LSP
 
 source:
-  id: pkg:github/zigtools/zls@0.10.0
-  asset:
-    - target: darwin_arm64
-      file: aarch64-macos.tar.zst
-      bin: bin/zls
-    - target: darwin_x64
-      file: x86_64-macos.tar.zst
-      bin: bin/zls
-    - target: linux_x64
-      file: x86_64-linux.tar.zst
-      bin: bin/zls
-    - target: linux_x86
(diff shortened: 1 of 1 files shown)


williamboman:
> * Is this the appropriate way to define/freeze a ZLS version? By specifying a commit hash (which was the latest commit at the time of this PR.)? If so, would we need to update the version every once-in-a-while?

This would be managed by Renovate. Commit hashes and building from source is generally discouraged because it targets unstable versions and, to your second question, relies on a proper build environment to be available.

I understand the zls package is close to unusable via Mason due to it being old and other complications related to compiler versions, but from the point of packaging it it's inherently a result of the upstream software being unstable. The latest version on Homebrew is 0.10.0, for example.

I'm not too keen on building from source like this for these reasons, especially without the necessary build harness to ensure successful builds.

PR #2379, opened 2023-08-07
initial add of mlir and tblgen lsp

zstreet87:
Hi there,

All three are from the LLVM project and are in the nvim-lspconfig repo so this should be straightforward.  However, this is my first time contributing so I may need some guidance.  In particular, please review the `source` and `bin`. 

https://github.com/neovim/nvim-lspconfig/blob/master/doc/server_configurations.md#mlir_lsp_server

https://github.com/neovim/nvim-lspconfig/blob/master/doc/server_configurations.md#mlir_pdll_lsp_server

https://github.com/neovim/nvim-lspconfig/blob/master/doc/server_configurations.md#tblgen_lsp_server

Attempting to resolve https://github.com/williamboman/mason-lspconfig.nvim/issues/234

### DIFF
--- a/packages/mlir_lsp_server/package.yaml
+++ b/packages/mlir_lsp_server/package.yaml
@@ -1,0 +1,16 @@
+---
+name: mlir-lsp-server
+description: The Language Server for the LLVM MLIR language
+homepage: https://mlir.llvm.org/docs/Tools/MLIRLSP/#mlir-lsp-language-server--mlir-lsp-server=
+licenses:
+  - Apache-2.0
+languages:
+  - Text
+categories:
+  - LSP
+
+source:
+  id: pkg:github/llvm/llvm-project@2023-8-07
+
+bin:
+  mlir-lsp-server: usr/bin/mlir-lsp-server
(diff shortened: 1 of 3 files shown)


williamboman (commented):
Hey! Thanks for the PR! The package definition is incomplete because it's not providing information about what to actually download, the only information is a GitHub repository.

This will be problematic to add as standalone packages because they are bundled with the entire clang+llvm bundle (which provides countless of binaries and is almost 1G large). For this reason I don't see it being feasible to add them to the registry, and users will have to install it elsewhere, for example installing llvm via the system manager.

zstreet87:
Oh I see; I thought there was a path since this is an option for Visual Studio

https://marketplace.visualstudio.com/items?itemName=llvm-vs-code-extensions.vscode-mlir

williamboman:
I believe the VSCode extension relies on these binaries to be installed externally in order to function. Neovim's equivalent to a dedicated VSCode language plugin would be a server configuration entry in `nvim-lspconfig`, or manually managed in your personal configuration via `:h lsp`.

Issue #2381, opened 2023-08-08
no matching distribution found for debugpy==1.6.8


leileimi:
when install debugpy ,error occurs 
ERROR: Could not find a version that satisfies the requirement debugpy==1.6.8 (from versions: 1.0.0b1, 1.0.0b2, 1.0.0b3, 1.0.0b4, 1.0.0b5, 1.0.0b6, 1.0.0b7, 1.0.0b8, 1.0.0b9, 1.0.0b10, 1.0.0b11, 1.0.0b12, 1.0.0rc1, 1.0.0rc2, 1.0.0, 1.1.0, 1.2.0, 1.2.1, 1.3.0, 1.4.0, 1.4.1, 1.4.2, 1.4.2.post1, 1.4.3, 1.5.0, 1.5.1, 1.6.0, 1.6.1, 1.6.2, 1.6.3, 1.6.4, 1.6.5, 1.6.6, 1.6.7)
        ERROR: No matching distribution found for debugpy==1.6.8
        spawn: /home/leiming/.local/share/lvim/mason/packages/debugpy/venv/bin/python failed with exit code 1 and signal 0.

MrGreenTea:
I've created a PR to revert the update #2344 

Until it get's merged you can do  `:MasonInstall debugpy@1.6.7` as a workaround and it should work (or at the least it did for me)

PR #2382, opened 2023-08-08
Revert wrong renovate update to debugpy

MrGreenTea:
The debugpy team tested their release pipeline and this new release triggered renovate (see PR ##2344) to update `debugpy` to v1.6.8 which does not exist on `pypi`. 

This fixes issue #2381

_Note: I would have liked to find a way to prevent renovate from doing this in the future, but didn't fully understand the configuration. To me it looks like you already use the `pypi` datasourceTemplate for this and therefore I don't exactly know if this can be prevented on `mason-registry`'s side._

### DIFF
--- a/packages/debugpy/package.yaml
+++ b/packages/debugpy/package.yaml
@@ -10,7 +10,7 @@ categories:
   - DAP
 
 source:
-  id: pkg:pypi/debugpy@1.6.8
+  id: pkg:pypi/debugpy@1.6.7
 
 bin:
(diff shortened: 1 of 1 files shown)


williamboman (approved):
Thanks! They probably just pulled that release from pypi. If Renovate runs before that happens, the CI will pass and the version will be upgraded. There's a setting in Renovate to wait X amount of time before merging version upgrades from package repositories that allow packages to pull versions, but it's not really been an issue so far.

PR #2425, opened 2023-08-11
feat: add kcl

starkers:
`Kcl` doesn't exist as a language in the json-schema yet, shall I add that to `schemas/enums/language.json` ?


williamboman (approved):
Thanks!

> Kcl doesn't exist as a language in the json-schema yet, shall I add that to schemas/enums/language.json ?

Yeah it'll need to be added. I wonder if "KCL" (uppercase) would be more correct?

PR #2431, opened 2023-08-12
chore(zls): update lsp schema location

### LINKED ISSUE #2430: Using Deprecated version of zls-vscode
The registry is currently using a deprecated version of [zls-vscode](https://github.com/zigtools/zls-vscode). It is recommended to use https://github.com/ziglang/vscode-zig instead.

https://github.com/mason-org/mason-registry/blob/66a9122a9f9213d6e4b9135e25e7841f5bb0ebff/packages/zls/package.yaml#L34-L35

that-ambuj:
Fixes #2430. This is supposed to improve DX and performance when using zls in neovim.

### DIFF
--- a/packages/zls/package.yaml
+++ b/packages/zls/package.yaml
@@ -32,7 +32,7 @@ source:
       bin: bin/zls.exe
 
 schemas:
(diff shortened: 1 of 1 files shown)


williamboman (approved):
Thanks! This has however nothing to do with performance, it's just the location of the configuration schema.

PR #2433, opened 2023-08-13
feat: add asmfmt

eeeXun:
Add asmfmt, formatter for assembly. Install with go.

### DIFF
--- a/packages/asmfmt/package.yaml
+++ b/packages/asmfmt/package.yaml
@@ -12,7 +12,7 @@ categories:
   - Formatter
 
 source:
-  id: pkg:golang/github.com/klauspost/asmfmt/cmd/asmfmt@v1.3.2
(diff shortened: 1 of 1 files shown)


packages/asmfmt/package.yaml
```diff
@@ -0,0 +1,18 @@
...
+homepage: https://github.com/klauspost/asmfmt
+licenses:
+  - MIT
+languages:
+  - Assembly
+categories:
+  - Formatter
+
+source:
+  id: pkg:golang/github.com/klauspost/asmfmt/cmd/asmfmt@v1.3.2
```
williamboman, line 15: For Renovate to successfully pick up new versions we need to separate the package name from the submodule (I forget the "Go" name for this).

```suggestion
  id: pkg:golang/github.com/klauspost/asmfmt@v1.3.2#cmd/asmfmt
```

williamboman (approved):
Thanks! I changed to use release assets since they provide that, much faster installations and removes dependency on Go.

PR #2436, opened 2023-08-13
bump(zls): update to v0.11.0

qazo:
Changed version number for the `zls` package to the latest (0.11.0). Latest package source archives no longer use zstd and are now packaged in gzipped tarballs and `.zip` files on Windows.

Also removed the `ci_skip` entries, as they were added there because the source archives where `.tar.zst`.

### DIFF
--- a/packages/zls/package.yaml
+++ b/packages/zls/package.yaml
@@ -39,12 +39,3 @@ schemas:
 
 bin:
   zls: "{{source.asset.bin}}"
-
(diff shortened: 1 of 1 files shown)


that-ambuj:
Thanks @qazo, I was about to open a similar PR. Turns out, You're faster than me.

williamboman (approved):
Thanks!

PR #2492, opened 2023-08-18
bump(nil): update to 2023-08-09

pimvh:
First of all, I am very happy with Mason. Thanks for this package!

I noticed that the Nil language for server is incompatible with newer versions of Nix. When building the version from Mason it fails. The fix incorporated in this newer version and should be backwards compatible. See oxalica/nil#93 for more context.

This also bumps to the latest version. There are more tags possible, I just picked the latest version of the language server. The tags are listed here: [oxalica/nil/tags](https://github.com/oxalica/nil/tags)

EDIT: I see I mistakenly tagged a PR in my commit message 🙃, I'll edit that on merge.

### DIFF
--- a/packages/nil/package.yaml
+++ b/packages/nil/package.yaml
@@ -12,7 +12,7 @@ categories:
 
 source:
   # renovate:versioning=loose
-  id: pkg:cargo/nil@2022-12-01?repository_url=https://github.com/oxalica/nil
+  id: pkg:cargo/nil@2023-08-09?repository_url=https://github.com/oxalica/nil
 
 bin:
   nil: cargo:nil


williamboman (approved):
Thanks! PRs are squash merged into a single commit, and I often update the commit message to follow the conventional commits format.

pimvh:
I think for test `Package tests / Test packages (darwin_x64)`, w might be dependant on this upstream fix (https://github.com/cachix/install-nix-action/issues/183). I just made a PR here to do a version bump mason-org/actions/pull/3.

With regards to the Linux tests (`Package tests / Test packages (linux_x64) (pull_request)` and `Package tests / Test packages (linux_x64_gnu) (pull_request)`), both give the same issue. It's unclear to me why this occurs, it might be related to how versioning is done in `nil` as in mason i always see it installing `v0.0.0`. Do you know the fix?

williamboman:
The version issue should be fixed in https://github.com/williamboman/mason.nvim/pull/1459

PR #2545, opened 2023-08-24
schemas: add support for swift based components

chibby0ne:
I want to add support for swift to mason-lspconfig, and noticed that the first step is to add the schema of swift based components.
Afterwards we can add the package entry for [sourcekit-lsp](https://github.com/apple/sourcekit-lsp) which is the LSP for swift.
And then the PR on the mason-lspconfig repo adding the mapping to the mason package.

This PR (hopefully) takes care of that first, step but since this is my first PR in this repository I'm not very familiar to all the ins and outs, I expect it to need a bit more changes.

Also I based the PURL on the ones used for golang since golang like swift makes use of packages. (Though I'm not very familiar with swift, in fact I just started learning it recently)

### DIFF
--- a/schemas/components/source.json
+++ b/schemas/components/source.json
@@ -14,7 +14,8 @@
                 { "$ref": "sources/npm" },
                 { "$ref": "sources/nuget" },
                 { "$ref": "sources/opam" },
-                { "$ref": "sources/pypi" }
+                { "$ref": "sources/pypi" },
+                { "$ref": "sources/swift" }
             ]
         }
     },
--- a/schemas/components/sources/swift.json
+++ b/schemas/components/sources/swift.json
@@ -1,0 +1,19 @@
+{
+    "$schema": "https://json-schema.org/draft/2020-12/schema",
+    "$id": "https://json.schemastore.org/mason-registry.json/components/sources/swift",
+    "type": "object",
+    "required": ["id"],
+    "properties": {
+        "id": {
+            "type": "string",
+            "pattern": "^pkg:swift/.+/.+@.+"
+        },
+        "extra_packages": {
+            "type": "array",
+            "description": "Extra swift packages required by the main package to function.",
+            "items": {
+                "type": "string"
+            }
+        }
+    }
+}
--- a/schemas/enums/language.json
+++ b/schemas/enums/language.json
@@ -145,6 +145,7 @@
         "Starlark",
         "Stylelint",
         "Svelte",
+        "Swift",
         "SystemVerilog",
         "TOML",
         "Teal",


williamboman (commented):
Hello, thanks for the PR! From a schema perspective this looks OK. However, how should "swift" packages be installed?

What does `pkg:swift/github.com/apple/sourcekit-lsp@swift-5.8.1-RELEASE` mean in terms of how a client (mason.nvim) should install it?

There's been requests to add the Sourcekit LSP before but my understanding has always been that it's not possible due to how it's distributed.

chibby0ne:
Hi @williamboman, thanks for the _swift_ reply (see what I did there :eyes:!)  

> What does pkg:swift/github.com/apple/sourcekit-lsp@swift-5.8.1-RELEASE mean in terms of how a client (mason.nvim) should install it?

I'm honestly I'm out of my depth here, but I initially thought that the PURL could be used to fetch the git tag and build it as a swift package (in case there's no prebuilt binary, which AFAIK there aren't any official ones unfortunately), the `swift:` would specify that it needs swift installed to build it :thinking: 

Is this not how mason works? Or does it fetch a binary and installs it in the right place (like go does when fetching packages or in this case downloading `gopls` from `goproxy.io` for example)?

> There's been requests to add the Sourcekit LSP before but my understanding has always been that it's not possible due to how it's distributed.

Do you refer to the fact that it is distributed as source and not a binary? I noticed that too and I also wasn't able to build it in linux following the instructions... :( 

On the other hand, how does vscode install sourcekit-lsp? :thinking: I know they require swift in the system but what happens then? Is it built from source? I haven't tried it but I assume it doesn't... it would take a loooong time if it does

williamboman:
> I'm honestly I'm out of my depth here, but I initially thought that the PURL could be used to fetch the git tag and build it as a swift package (in case there's no prebuilt binary, which AFAIK there aren't any official ones unfortunately), the `swift:` would specify that it needs swift installed to build it 🤔

I certainly don't know, I'm not a Swift developer. I'd have assumed there's a package manager or similar for Swift.

> Do you refer to the fact that it is distributed as source and not a binary? I noticed that too and I also wasn't able to build it in linux following the instructions... :(

Mostly the fact that it seems highly coupled with the Swift toolchain and it comes bundled with the Swift toolchain already, I'm not sure adding it to mason would be a good idea even if it was possible.

jameschensmith:
> On the other hand, how does vscode install sourcekit-lsp? 🤔 I know they require swift in the system but what happens then? Is it built from source? I haven't tried it but I assume it doesn't... it would take a loooong time if it does

If you're referring to [swift-server/vscode-swift](https://github.com/swift-server/vscode-swift), they don't install SourceKit-LSP. [They default the path to the one that comes with the Swift toolchain](https://github.com/swift-server/vscode-swift/blob/2f29b2b916735a04d0b2a7d105d5946e89f5563e/docs/settings.md?plain=1#L45-L47), as William noted. Are there any other languages that bundle their LSP with the toolchain? I will say Apple's [docs for using SourceKit-LSP with Vim / Neovim](https://github.com/apple/sourcekit-lsp/tree/f32e7dd51cc72bffeeedc6c4750e50ad6769e866/Editors#vim-8-or-neovim) seems to cover how to set it up with `lspconfig` pretty well. I'm not sure if much more can be done 🤔

PR #2563, opened 2023-08-26
feat: add perl-debug-adapter

WhoIsSethDaniel:


WhoIsSethDaniel:
This will allow me to install `perl-debug-adapter` (aka vscode-perl-debug) locally when I use a file: registry. I'm far from a javascript/typescript person so there may well be better ways to compile this code.

packages/perl-debug-adapter/package.yaml
```diff
@@ -0,0 +1,20 @@
+---
+name: perl-debug-adapter
+description: Perl debugger
```
williamboman, line 3: Figured this might be bit more descriptive 😆 
```suggestion
description: Perl debugger that invokes perl -d.
```

packages/perl-debug-adapter/package.yaml
```diff
@@ -0,0 +1,20 @@
...
+homepage: https://github.com/Nihilus118/vscode-perl-debug
+licenses:
+  - MIT
+languages:
+  - Perl
+categories:
+  - DAP
+
+source:
+  id: pkg:github/Nihilus118/vscode-perl-debug@7ef298610b91b76ed76bebc4962f1a849052a437
```
williamboman, line 13: It'd be ideal if they tagged releases, seeing as they do version it. Even better would be if they packaged this as a npm package. Feel free to ask them about this ;)

```suggestion
  # renovate:datasource=git-refs
  id: pkg:github/Nihilus118/vscode-perl-debug@7ef298610b91b76ed76bebc4962f1a849052a437
```

williamboman (approved):
Thanks!

packages/perl-debug-adapter/package.yaml
```diff
@@ -0,0 +1,21 @@
...
+categories:
+  - DAP
+
+source:
+  # renovate:datasource=git-refs
+  id: pkg:github/Nihilus118/vscode-perl-debug@7ef298610b91b76ed76bebc4962f1a849052a437
+  build:
+    run: |
+      npm install
+      npx tsc
```
williamboman, line 18: This should remove dependencies only required to build the package. Also changing to use their `compile` script in case it changes over time.

```suggestion
      npm install
      npm run compile
      npm install --production
```

PR #2567, opened 2023-08-26
feat: add snyk-ls

Dueen:
Adds the Snyk language server
I have not been able to test this locally and would appreciate some help in how to do so


williamboman (approved):
Thanks!

> I have not been able to test this locally and would appreciate some help in how to do so

Yeah this is very tricky currently, but I'm just about to add some docs on how to do this easily locally.

PR #2640, opened 2023-09-03
feat: add `stylelint` addons and `bibtex-tidy`

chrisgrieser:
- `stylelint` is commonly used with a config (since it has no rules enabled by default) and the `stylelint-order` plugin. Without having them in mason, too, mason's `stylelint` often cannot be used. I chose to add the most common config (recommended by stylelint in their getting started guide), and the by far most common plugin (with config), to not open the gate to add every negligible plugin of a linter.
- Also, I added `bibtex-tidy`. It's actually a formatter for `bibtex` files, but the schema only accepts `latex`, so that's what I filled in for now to conform with the schema.

### DIFF
--- a/packages/bibtex-tidy/package.yaml
+++ b/packages/bibtex-tidy/package.yaml
@@ -1,0 +1,16 @@
+---
+name: stylelint-config-standard
+description: The standard shareable config for Stylelint
+homepage: https://github.com/stylelint/stylelint-config-standard
+licenses:
+  - MIT
+languages:
+  - CSS
+  - Sass
+  - SCSS
+  - LESS
+categories:
+  - Linter
+
+source:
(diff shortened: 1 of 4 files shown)


williamboman (commented):
Hey! I believe these are plugins to stylelint itself? In that case they'll need to be installed in a way that is compatible with Node's module resolution algorithm, i.e. in a hierarchical manner walking up the fs. By having these as separate packages they will be placed adjacent to one another which won't work. See https://github.com/williamboman/mason.nvim/issues/392.

chrisgrieser:
I see. Yes that is the case, so the stylelint packages will have to be deferred. 

Bibtex tidy is not be affected though. I'll open a new PR for that

PR #2652, opened 2023-09-04
feat: add `bibtex-tidy`

chrisgrieser:
follow up to the closed #2640.

This one includes only the `bibtex-tidy` package. It's a formatter for `bibtex`, but the yaml schema does not allow that as language and only accepts`LaTeX` as language, so that's what I set the language to. (Assuming tests would fail if I set it to `bibtex`.)


williamboman:
@williambotman /fixup

williamboman:
Ah the `.yaml` was fixed already, seems like the bot picked up some other files instead 🤷. Thanks!

chrisgrieser:
Okay, is there anything else to do here?

PR #2764, opened 2023-09-14
feat: add llm-language-server package

wilfriedroset:


williamboman (commented):
Thanks! I wonder if `llm-ls` would be a better name, as it would match the binary name and the upstream repo name, according to Mason's naming standards.

PR #2807, opened 2023-09-18
feat: add blueprint

FineFindus:
Adds the [Blueprint Language Server](https://gitlab.gnome.org/jwestman/blueprint-compiler), closing https://github.com/williamboman/mason.nvim/issues/412.
This is marked as a draft, as I neither tested the code locally, nor does it actually work.
The Blueprint language server is distributed in the same binary as the compiler, and can be started with the `--lsp` flag. Is there any way to add a flag to the executed binary? The [Contributing guide](https://github.com/mason-org/mason-registry/blob/main/CONTRIBUTING.md) did not mention any.

### DIFF
--- a/packages/blueprint/package.yaml
+++ b/packages/blueprint/package.yaml
@@ -1,0 +1,28 @@
+---
+name: blueprint_ls
+description: Language server for Blueprint, a markup language and compiler for GTK 4 user interfaces.
+homepage: https://gitlab.gnome.org/jwestman/blueprint-compiler
+licenses:
+  - LGPL-3.0-or-later
+languages:
+  - Blueprint
+categories:
+  - LSP
+
+source:
+  id: pkg:generic/blueprint?vcs_url=git%2Bhttps://gitlab.gnome.org/jwestman/blueprint-compiler%402a39a16391122af2f3d812e478c1c1398c98b972
+  build:
+    - target: unix
+      run: |
+        meson -Dprefix="$PWD" build
+        ninja -C build install
+      bin: bin/blueprint-compiler
+    - target: win
+      run: |
+        meson -Dprefix="($pwd).path" build
+        ninja -C build install
+      bin: bin/blueprint-compiler.exe
+
+
+bin:
+    blueprint-compiler: "{{source.build.bin}}"
--- a/schemas/enums/language.json
+++ b/schemas/enums/language.json
@@ -20,6 +20,7 @@
         "Beancount",
         "Bicep",
         "Blade",
+        "Blueprint",
         "BrighterScript",
         "C",
         "C#",


williamboman:
> Is there any way to add a flag to the executed binary? The [Contributing guide](https://github.com/mason-org/mason-registry/blob/main/CONTRIBUTING.md) did not mention any.

Mason is not involved in starting LSP servers. Such information would go into for example [nvim-lspconfig](https://github.com/neovim/nvim-lspconfig/blob/master/lua/lspconfig/server_configurations/blueprint_ls.lua). Mason only makes the executable available, it doesn't concern itself with how it's invoked.

packages/blueprint/package.yaml
```diff
@@ -0,0 +1,28 @@
...
+homepage: https://gitlab.gnome.org/jwestman/blueprint-compiler
+licenses:
+  - LGPL-3.0-or-later
+languages:
+  - Blueprint
+categories:
+  - LSP
+
+source:
+  id: pkg:generic/blueprint?vcs_url=git%2Bhttps://gitlab.gnome.org/jwestman/blueprint-compiler%402a39a16391122af2f3d812e478c1c1398c98b972
```
williamboman, line 13: The mason.nvim client would currently not be able to parse this PURL.

The purl spec includes a `gitlab` pkg type, which I think would be more fitting here. However, mason.nvim currently doesn't support this package type, so that would have to be added first.

```suggestion
  id: pkg:gitlab/jwestman/blueprint-compiler@402a39a16391122af2f3d812e478c1c1398c98b972
```

williamboman: Oh it's a self-hosted instance, that complicates things. The generic type might be better.

FineFindus: I saw the type, but was not sure if it would work since the repository is not hosted on the gitlab.com instance. Will it still work?
What is required for mason.nvim to support the package type?

williamboman (commented):
Thanks! I can't find any docs on how to install this, and their docs mention it's experimental. Also as I mentioned in another comment, for this package to be installable by the mason.nvim client support will have to be added there first.

FineFindus:
It can be installed like most meson projects using `meson _build` and `ninja -C _build install`, it's a bit hidden on the [setup page](https://jwestman.pages.gitlab.gnome.org/blueprint-compiler/setup.html). While it is still experimental, it is already being used in quite a few GTK/GNOME applications, with growing adoption.
What needs to be done to make the package installable from mason.nvim? I'm happy to add support for it if needed.

FineFindus:
Just a friendly ping to check the status of this pr :)
I'm still willing to add support for the required purl spec if I get a few pointers on how to get started. Alternatively, I could set up a GitHub mirror of the project so we can download releases from there.

CleoMenezesJr:
Cool. I hope it will be available soon

williamboman:
I know this has been open for a very long time, but in order to reach some conclusion I'll go ahead and close this with the following reason: Custom build scripts are being deprecated because of difficulty in maintaining them over time and their common reliance on system dependencies that Mason cannot detect. You can always host your own registry if you'd like to manage it through Mason, see https://github.com/williamboman/mason-registry-playground/ for a starting point

PR #2837, opened 2023-09-21
Remove Java dependency from ltex-ls on Windows

scottmckendry:
Mason currently uses the platform-independent release for ltex-ls. As per [documentation](https://valentjn.github.io/ltex/ltex-ls/installation.html) this is not the recommended approach:

> It’s recommended that you choose the archive corresponding to your platform (these archives are standalone, no Java installation necessary).
If you choose the platform-independent file ltex-ls-VERSION.tar.gz, then you need Java 11 or later on your computer.

I've only updated the Windows asset but would also be happy to do the same testing on a Linux machine and open another PR if this is approved and merged.

One thing I noticed when testing this is I had to update the default cmd (`ltex-ls`) to the following in my config:

```lua
    -- Ltex LS (LanguageTool)
    local ltex_cmd = vim.fn.stdpath("data") .. "/mason/packages/ltex-ls/ltex-ls-16.0.0/bin/ltex-ls"
    require("lspconfig").ltex.setup({
      cmd = { ltex_cmd },
      settings = {
        ltex = {
          language = "en-GB",
        }
      }
    })
```

Not sure if this is a mason-lspconfig issue or if it's because I've pointed mason at my forked registry. Just mentioning in case this causes problems for others. 

Apart from that, everything else appears to be working as expected.

### DIFF
--- a/README.md
+++ b/README.md
@@ -13,4 +13,3 @@
 Refer to [CONTRIBUTING.md](./CONTRIBUTING.md) for contribution guidelines.
 
 Compiled registry contents are available via [releases](https://github.com/mason-org/mason-registry/releases).
-
--- a/README.md
+++ b/README.md
@@ -13,3 +13,4 @@
 Refer to [CONTRIBUTING.md](./CONTRIBUTING.md) for contribution guidelines.
 
 Compiled registry contents are available via [releases](https://github.com/mason-org/mason-registry/releases).
+
--- a/packages/ltex-ls/package.yaml
+++ b/packages/ltex-ls/package.yaml
@@ -21,7 +21,7 @@ source:
         lsp: ltex-ls-{{version}}/bin/ltex-ls
         cli: ltex-ls-{{version}}/bin/ltex-cli
     - target: win
-      file: ltex-ls-{{version}}.tar.gz
+      file: ltex-ls-{{version}}-windows-x64.zip
       bin:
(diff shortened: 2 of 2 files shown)


williamboman (commented):
Thanks! I remember looking into this a few times before actually. I believe the reason it's not using the standalone variants is 1) because they only provide x64 targets for Mac, Linux/GNU & Windows and 2) it'd add quite a significant filesystem footprint by vendoring a standalone JRE.

This would only really be problematic for Linux users as it would exclude arm64 and non-GNU distributions. Mac arm64 systems would only be penalized with a minor overhead from the Rosetta x64 -> arm64 translation, and I'd wager 99%+ of Windows users are x64.

It's difficult to make an informed decision as there's no usage data, but generally anything that would exclude Linux arm64 is a no-go, and Mac Rosetta translation is acceptable if there are no alternatives. I understand the problem of `ltex-ls` currently having an implicit dependency on JRE11+ being installed on the system but I'm not sure reducing the amount of supported platforms and vendoring JRE is the right approach.

scottmckendry:
@williamboman completely understand. I'm definitely biased towards not having to install an external dependency if I can avoid it. 

But I agree that compatibility should be the first priority so will close this one off for now.

williamboman:
@scottmckendry Just FYI this might end up being of interest after all, see https://github.com/williamboman/mason.nvim/issues/1531.

PR #2845, opened 2023-09-21
feat: add tsserver

perrin4869:
follow-up on https://github.com/pmizio/typescript-tools.nvim/pull/152
should facilitate adding `tsserver` to `PATH` and solve the issues in that PR

### DIFF
--- a/packages/tsserver/package.yaml
+++ b/packages/tsserver/package.yaml
@@ -1,0 +1,17 @@
+---
+name: tsserver
+description: The TypeScript standalone server.
+homepage: https://github.com/microsoft/TypeScript/wiki/Standalone-Server-(tsserver)
+licenses:
+  - MIT
+languages:
+  - TypeScript
+  - JavaScript
+categories:
+  - LSP
+
+source:
+  id: pkg:npm/typescript@5.2.2
+
+bin:
+  tsserver: npm:tsserver


perrin4869:
@williamboman just wondering, is there anything blocking this?

williamboman:
> @williamboman just wondering, is there anything blocking this?

Yeah I was wondering about the implications of linking the `tsserver` via Mason due to how the executable resolution was done in typescript-tools.nvim. It seems like the PR was merged now and uses the `typescript-language-server` package - is there still a use case for a separate package? I think it would make sense because `typescript` is an opaque dependency of `typescript-language-server` whose version you cannot control, and there might come a time when that dependency is removed.

I saw in the PR in typescript-tools.nvim it also reaches into the `packages/` directory. This is also prone to breakage because that "API" is not stable and is subject to change. For example I'm planning on enabling multi-version installations, which might result in a file tree hierarchy like so:

```
packages/
| packages/typescript-language-server/
  | packages/typescript-language-server/3.3.2/
  | packages/typescript-language-server/2.0.2/
```

Instead, files that are to be accessed elsewhere should instead be linked via `bin` (executables), `share` (platform-agnostic files), or `opt` (platform-specific files), so should this package be added I think we should link contents into opt, like so:

```yaml
opt:
  tsserver: ./node_modules/typescript
```

perrin4869:
Yeah, I look at the typescript-tools.nvim PR as maybe a way to attempt to provide support to users who only have `typescript-language-server` available via mason, but yeah, I guess the resolution method may still be fragile...

I was wondering what the `opt` in the path in your comment on that PR was 😅
Well, typescript-tools will still prefer using tsserver via PATH if it's available, and that's what this package provides, and that's a more stable way to use it?

perrin4869:
@pmizio was wondering, do you have any input regarding this?

PR #3003, opened 2023-10-06
feat: add gci

fredrikaverpil:
https://github.com/daixiang0/gci
https://pkg.go.dev/github.com/daixiang0/gci


williamboman (approved):
Thanks!

PR #3083, opened 2023-10-14
Build java-test directly rather than depending on release asset

### LINKED ISSUE #3036: Updating java-test no longer possible
Unfortunately java-test does not provide `.vsix` files on their release page since v0.39.0 which is what the [current package-specification](https://github.com/mason-org/mason-registry/blob/main/packages/java-test/package.yaml) depends on.

This is probably also why [this](https://github.com/microsoft/vscode-java-test/releases) failed.

A workaround would be to build the package from source, [which seems to be mostly done with npm](https://github.com/microsoft/vscode-java-test/blob/main/.github/workflows/build.yml).

I would be happy to contribute if I could get some pointers on testing the changes to the registry locally.

Is there a way to prevent mason from fetching the registry remotely and somehow point it to a locally pulled registry which I can use for development?

brianrodri:
Fixes #3036

Mason can't update to `vscode-java-test > 0.39.0` because `vscjava.vscode-java-test-XX.vsix` is no longer included in the release assets.

[microsoft/vscode-java-test#1594](https://github.com/microsoft/vscode-java-test/issues/1594#issuecomment-1670522559) instructs users to build it themselves by following their GitHub Workflow script. Specifically, the command [`npm run build-plugin`](https://github.com/microsoft/vscode-java-test/blob/main/.github/workflows/build.yml#L40-L41) generates all of the `.jar` files in the `server/` directory.

Unfortunately, this directory is different than `extension/server/`. This would break pretty much every plugin depending on the location of those `.jar` files.

I can only think to add additional `build.run` steps to move to the expected directory in a platform-independent way. I'm not sure if that's desired/necessary, or maybe there's an existing way to correct this behavior? I tried exporting to the correct `share/` directory, but the plugin I've been testing with, LazyVim, depends on the install location not the share location.

Would appreciate some guidance here!

### DIFF
--- a/packages/java-test/package.yaml
+++ b/packages/java-test/package.yaml
@@ -21,9 +21,11 @@ categories:
   - DAP
 
 source:
-  id: pkg:github/microsoft/vscode-java-test@0.39.0
-  asset:
-    file: vscjava.vscode-java-test-{{version}}.vsix
+  id: pkg:github/microsoft/vscode-java-test@0.40.0
+  build:
+    run: |
+      npm install
+      npm run build-plugin
 
 share:
-  java-test/: extension/server/
+  java-test/extension/server/: server/


nsoufian:
Any updates on this ?

williamboman:
Thanks for the PR! This introduces a number of external dependencies (node & npm being the most obvious ones). I'm not sure what the `build-plugin` script does entirely but I believe it simply builds things via maven. Microsoft seems to have removed assets from multiple of their projects, so java-test is not the only package that is unable to upgrade currently. I've been meaning to host built assets for these projects, probably through something similar to https://github.com/mason-org/julia-lsp. Surfacing these build commands directly to users will definitely cause issues and is a significant source of issue reports which is why I'm not keen on merging this as-is.

As for the share files, as long as they are in the same location as before within the `share/` directory it would not be a breaking change. I believe it should be:

```yaml
share:
  java-test/: server/
```

williamboman:
In the meantime, if you want to access this particular version of the package definition so that you can upgrade it, you can clone @brianrodri's fork and check out the `fix-java-test` branch and then configure Mason to source packages locally from your file system:

```lua
require("mason").setup {
  registries = {
    "file:~/path/to/mason-registry"
  }
}
```

Another option would be to add this `package.yaml` to your Neovim config by placing it in `~/.config/nvim/packages/java-test/package.yaml` and then configuring Mason like so:

```lua
require("mason").setup {
  registries = {
    "file:" .. vim.fn.stdpath("config"),
    "github:mason-org/mason-registry",
  }
}
```

This way you can host your own package overrides inside your own Neovim config and version them accordingly.

zenoli:
@williamboman I tried your second proposed workaround but I get:
`mason.nvim  ERROR Unknown registry type "file": "file:/home/xxxxx/.config/nvim"`

Looking at the mason code, it seems that only the types "github" and "lua" are supported.
Is there a way to add a local registry as you suggested?

I think this would generally be very useful, especially if someone wants to contribute to the registry and test changes locally first. Or how would one normally test changes other than forking the registry?

s1n7ax:
Hello. I have created a workflow that release build artifacts on new release of java-test. This is daily checking to see if there are new relea are available in the original project that's not available in the current releases. 

NOTE that `0.40.1` release fail to build so at the moment I have used main branch to release the build artifacts.

https://github.com/nvim-java/vscode-java-test-releases

I have a custom mason registry to use the releases from this here

https://github.com/nvim-java/mason-registry

Adding the registry to mason & `Mason` should to the job. `MasonInstall java-test@0.40.1` should install the newest version.

```lua
{
	'williamboman/mason.nvim',
	opts = {
		registries = {
			'github:nvim-java/mason-registry',
			'github:mason-org/mason-registry',
		},
	},
}
```

williamboman:
See https://github.com/mason-org/mason-registry/pull/3348#issuecomment-1871736659

PR #3113, opened 2023-10-17
feat: add woke

chrisgrieser:
First time I am adding a registry that is not simple copypaste from a similar registry. Kudos for the detailed guidelines, made it much easier.

The release page for reference: https://github.com/get-woke/woke/releases

### DIFF
--- a/packages/woke/package.yaml
+++ b/packages/woke/package.yaml
@@ -12,28 +12,28 @@ source:
   asset:
     - target: darwin_arm64
       file: woke-{{ version | strip_prefix "v" }}-darwin-arm64.tar.gz
-      bin: woke
+      bin: woke-{{ version | strip_prefix "v" }}-darwin-arm64/woke
     - target: darwin_x64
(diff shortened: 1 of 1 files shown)


chrisgrieser:
Okay, so I managed to figure out why the tests were failing for the most part, this round of test failures I simply cannot figure out.

the relevant error seems to be this:
```log
[mason] [ERROR 19:19:07] ...ns/v1/tests/mason.nvim/lua/mason-core/installer/init.lua:249: Installation failed for Package(name=woke) error='Tried to link bin "woke" to non-existent target "woke".'
```

However, there is exactly a binary named "woke" at the package root, so am not really sure what the remaining issue here is.

chrisgrieser:
Nevermind, I figured it out. The problem was that my unarchiving app was a little bit too smart, as it automatically unnests files when there is only one folder in a folder, effectively hiding from me that the `woke` binaries were nested one level deeper 🙈

williamboman (approved):
Thanks!

PR #3117, opened 2023-10-18
feat: add cairo-language-server

j1mbo64:


williamboman (approved):
Thanks! Using their release assets would be preferable but they only provide for Linux x64 and Mac arm64.

PR #3139, opened 2023-10-19
Add ELP (new Erlang language server)

filmor:


williamboman (approved):
Thanks! A+ for including LSP schema entry, renovate versioning comment, and `_gnu` target!

PR #2804, opened 2023-09-18
add easy-coding-standard

fbuchlak:

PR #3225, opened 2023-10-26
add easy-coding-standard

fbuchlak:
Sorry for bumping, I've accidentally closed #2804 

There was a [problem](https://github.com/mason-org/mason-registry/actions/runs/6240396709/attempts/2) with test for win x64

Only two packages in registry use composer so far
`phpactor` - has only support for unix
`psalm` having same problem (see #1209) as `easy-coding-standard`


williamboman (approved):
> Only two packages in registry use composer so far `phpactor` - has only support for unix `psalm` having same problem (see #1209) as `easy-coding-standard`

Ah nice investigation! Not sure why it's freezing on Windows. Thanks!

PR #3239, opened 2023-10-27
feat: add zprint

fbuchlak:


packages/zprint/package.yaml
```diff
@@ -0,0 +1,24 @@
...
+  - Formatter
+
+source:
+  id: pkg:github/kkinnear/zprint@1.2.8
+  asset:
+    - target: darwin_arm64
+      file: zprintma-{{version}}
+    - target: darwin_x64
+      file: zprintm-{{version}}
+    - target: linux_x64
```
williamboman, line 20: This is a GNU binary so we'll need to add an additional constraint

```suggestion
    - target: linux_x64_gnu
```

williamboman (approved):

PR #3240, opened 2023-10-27
feat: add zprint-clj

fbuchlak:
add zprint-clj (js wrapper) for #3239
might help windows users

### DIFF
--- a/packages/zprint-clj/package.yaml
+++ b/packages/zprint-clj/package.yaml
@@ -1,0 +1,17 @@
+---
+name: zprint-clj
+description: Node.js wrapper for ZPrint Clojure source code formatter
+homepage: https://github.com/clj-commons/zprint-clj
+licenses:
+  - MIT
+languages:
+  - Clojure
+  - ClojureScript
+categories:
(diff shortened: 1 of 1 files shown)


williamboman (commented):
Thanks! Is this project maintained? Last activity ~3 years ago

fbuchlak:
> Thanks! Is this project maintained? Last activity ~3 years ago

Probably not. I use `zprint` personally, which references this wrapper in it's [documentation](https://github.com/kkinnear/zprint/blob/main/doc/getting/other.md). Wrapper uses [zprint 0.5.3](https://github.com/clj-commons/zprint-clj/blob/master/deps.edn) (current version is 1.2.8).
I've added this only because zprint does not ship windows build. Consider closing this PR, however I've tested this locally, and it could produce decent formatting, so it might be useful for someone else.

williamboman:
> I've added this only because zprint does not ship windows build. Consider closing this PR, however I've tested this locally, and it could produce decent formatting, so it might be useful for someone else.

This seems like a good reason to include it.

Issue #3236, opened 2023-10-27
CI - win_x64 test fails for composer


fbuchlak:
Failing test for win_x64 when using `source.id: pkg:composer/*`

Related to:
#1209
#3225

PR #3243, opened 2023-10-27
feat: add phpinsights

fbuchlak:
adds phpinsights

skip test for win_x64 - see #3236


williamboman (approved):
This is failing currently because of:

```
dealerdirect/phpcodesniffer-composer-installer contains a Composer plugin which is blocked by your allow-plugins config. You may add it to the list if you consider it safe.
You can run "composer config --no-plugins allow-plugins.dealerdirect/phpcodesniffer-composer-installer [true|false]" to enable it (true) or disable it explicitly and suppress this exception (false)
See https://getcomposer.org/allow-plugins
```

I can't really reproduce this on the couple of machines I've tried on, so I'm not sure if the GitHub runner that comes preinstalled with Composer & PHP is configured differently than the defaults..

If the runner fails with this error I suspect some percentage of users will also run into this, I'm not sure if this warning is actually benign or if it's just a matter of configuration.

If possible, we could solve this in the mason.nvim client directly by allowing plugins (locally in Mason's installation directory only). This could either be default behavior or explicitly enabled via the PURL specifier, e.g. ` pkg:composer/nunomaduro/phpinsights@v2.9.0?allow_plugins=true`.

fbuchlak:
> If possible, we could solve this in the mason.nvim client directly by allowing plugins (locally in Mason's installation directory only). This could either be default behavior or explicitly enabled via the PURL specifier, e.g. `pkg:composer/nunomaduro/phpinsights@v2.9.0?allow_plugins=true`.

I would rather go with explicitly enabling plugins, but I'm not sure if we want to enable all plugins.

It would be nice to have something similar to `source.build`
```yaml
source:
  id: pkg:composer/acme/package@v1.13.5
  init: 
    run: |
      composer config allow-plugins.acme/plugin-1 true
      composer config allow-plugins.acme/plugin-2 true
      composer config allow-plugins.acme/plugin-3 true
```
or just `source.allowed_plugins` for composer.

However we don't need to do this at all. This package can be installed same as [phpactor](https://github.com/mason-org/mason-registry/blob/main/packages/phpactor/package.yaml) (github source)

PR #3256, opened 2023-10-28
feat(trivy): add trivy linter

pbnj:


packages/trivy/package.yaml
```diff
@@ -0,0 +1,50 @@
+---
+name: trivy
+description: A comprehensive and versatile security scanner
```
williamboman, line 3: Might as well use the description from their repository which is a bit more descriptive imo
```suggestion
description: |
  Find vulnerabilities, misconfigurations, secrets, SBOM in containers, Kubernetes, code repositories, clouds and more.
```

pbnj:
Thank you @williamboman

PR #3286, opened 2023-10-31
Add dbt templater to SQLfluff

mlavaert:
SQLFluff is often used to format projects that use DBT. This templater used to be part of SQLFluff but now it has been extracted to a separate package.

> From sqlfluff version 0.7.0 onwards, the dbt templater has been moved to a separate plugin and python package.
Projects that were already using the dbt templater may initially fail after an upgrade to 0.7.0+. See the installation
instructions below to install the dbt templater.

Re-adding it here allows users to format their DBT projects again.

### DIFF
--- a/packages/ast-grep/package.yaml
+++ b/packages/ast-grep/package.yaml
@@ -28,7 +28,7 @@ categories:
   - LSP
 
 source:
-  id: pkg:npm/%40ast-grep/cli@0.19.0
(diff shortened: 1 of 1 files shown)


williamboman (commented):
Thanks! I'm in the midst of adding support for [subpackages](https://github.com/williamboman/mason.nvim/issues/392) which I think would suit this better. I just need to finalize some docs and clean up some commit histories before it's ready (some time later this week)

PR #3332, opened 2023-11-04
feat: add ast-grep

chrisgrieser:
can be used as linter, & it's used by telescope extensions for searching


packages/ast-grep/package.yaml
```diff
@@ -0,0 +1,33 @@
...
+  - Dart
+  - Lua
+
+categories:
+  - Linter
+  - Formatter
+  - Runtime
+
+source:
+  id: pkg:npm/@ast-grep/cli@0.12.5
```
williamboman, line 30:
```suggestion
  id: pkg:npm/%40ast-grep/cli@0.12.5
```

williamboman:
They seem to provide an `sg` executable as well, and in their docs it seems like it's the preferred one. Any reason not to include it or should we add it as well?

chrisgrieser:
nope, just haven't thought of including both. Added `sg`

PR #3348, opened 2023-11-06
bump up java-test to 0.40.1

s1n7ax:
After the removal of .vsix from releases, updating `java-test` is not possible. So, this change will build the package from source. https://github.com/mason-org/mason-registry/issues/3036

IMPORTANT: path to shared `jar`s  in the `vsix` was `java-test/extension/server/*.jar` but when build with source, there is no `extension` directory anymore. This might break existing configs.

Current latest version in Mason-registry also has a bug where there is an error upon test search. https://github.com/microsoft/vscode-java-test/issues/1625 so updating current version is important.

### DIFF
--- a/packages/java-test/package.yaml
+++ b/packages/java-test/package.yaml
@@ -28,4 +28,4 @@ source:
       npm run build-plugin
 
 share:
-  java-test/: extension/server/
+  java-test/: server/
--- a/packages/java-test/package.yaml
+++ b/packages/java-test/package.yaml
@@ -21,9 +21,11 @@ categories:
   - DAP
 
 source:
-  id: pkg:github/microsoft/vscode-java-test@0.39.0
-  asset:
-    file: vscjava.vscode-java-test-{{version}}.vsix
+  id: pkg:github/microsoft/vscode-java-test@0.40.1
+  build:
+    run: |
+      npm install
+      npm run build-plugin
 
 share:
   java-test/: extension/server/


mikehaertl:
@williamboman I'd like to help testing this as this update is quite important for Java Devs.

Can you (or @s1n7ax ) give me a quick pointer how configure mason to point to the source branch of this PR?

I found this, but unsure how to set the branch:
https://github.com/williamboman/mason.nvim/blob/41e75af1f578e55ba050c863587cffde3556ffa6/README.md?plain=1#L221-L227

This information could probably also be added to CONTRIBUTING.md

s1n7ax:
@mikehaertl 0.40.1 clean build stopped working for me. I don't know if it worked the first time because I had a cache. Main branch however is working. So, I forked the [vscode-java-test](https://github.com/nvim-java/vscode-java-test) and created a new tag 0.40.2 from main. I created a new mason registry called [nvim-java-mason-registry](https://github.com/nvim-java/nvim-java-mason-registry). It uses my java-test fork and `0.40.2` tag. If you want to use that, you can add the registry to mason config. After adding the registry, you should run `MasonUpdate`. 

```lua
return {
	'williamboman/mason.nvim',
	opts = {
		registries = {
			'github:nvim-java/nvim-java-mason-registry',
			'github:mason-org/mason-registry',
		},
	},
}
```

mikehaertl:
@s1n7ax Thanks for the update and for the pointer to that other issue. 

I quickly tried your registry but had some trouble due to changed paths or something and had no time to fix my config. TBH I will then probably wait a bit and see if there's an "official" solution.

Thanks for your effort nonetheless.

s1n7ax:
Hello. I have created a workflow that release build artifacts on new release of java-test. This is daily checking to see if there are new relea are available in the original project that's not available in the current releases. 

NOTE that `0.40.1` release fail to build so at the moment I have used main branch to release the build artifacts.

https://github.com/nvim-java/vscode-java-test-releases

I have a custom mason registry to use the releases from this here

https://github.com/nvim-java/mason-registry

Adding the registry to mason & `Mason` should to the job. `MasonInstall java-test@0.40.1` should install the newest version.

```lua
{
	'williamboman/mason.nvim',
	opts = {
		registries = {
			'github:nvim-java/mason-registry',
			'github:mason-org/mason-registry',
		},
	},
}
```

williamboman:
Nice @s1n7ax! I like the idea of the community taking ownership of specific areas :+1: (I know creating your own Mason registry is a hassle currently, I need to make it much easier and document it).

As for `java-test` and `java-debug`, the good news is that [OpenVSX](https://open-vsx.org/) seems to have been revitalized earlier this year after announcing they were shutting down their services. They host built artifacts for both `java-test` and `java-debug`, and I'm right now working on adding OpenVSX support in the mason.nvim client. Once that's done we can update these packages to install from OpenVSX (e.g. `pkg:openvsx/vscjava/vscode-java-test@0.40.1`). I'll close this PR and create new ones once it's ready to go.

s1n7ax:
@williamboman That's better. Idea of adding third party registries are brilliant however, in reality, pretty much most issues we got had something to do with updating mason registries. So far it's always user messing up something when setting up mason. I would like some improvements when it comes to mason APIs available for plugin developers at least to show better error messages to user.

I will create an issue with challenges but in simple words, registry versioning & dynamic registry loading would be perfect.

PR #3390, opened 2023-11-08
docs(custom-elements-ls): Update docs, version and languagelist

Matsuuu:
Heya, Author of Custom Elements Language Server here. Updating the filetypes of the plugin to support all files. Also cleaning up some of the docs around installation

### DIFF
--- a/packages/custom-elements-languageserver/package.yaml
+++ b/packages/custom-elements-languageserver/package.yaml
@@ -1,24 +1,15 @@
 ---
 name: custom-elements-languageserver
 description: |
-  The best way to utilize the Custom Elements Language Server is to enable the Custom Elements Manifest (CEM) in your
-  project by installing a CEM generator like one provided by The Open WC Team.
-
-  Generating a CEM in watch mode will provide you with the best user experience. If your dependencies ship with a Custom
-  Elements Manifest, those will be utilized also.
+  Custom Elements Language Server provides useful language features for Web Components. Features include code actions, completons, diagnostics and more.
 homepage: https://github.com/Matsuuu/custom-elements-language-server
 licenses:
   - BSD-3-Clause
-languages:
-  - HTML
-  - JavaScript
-  - TypeScript
-  - JSX
 categories:
   - LSP
 
 source:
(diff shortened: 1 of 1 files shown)


packages/custom-elements-languageserver/package.yaml
```diff
@@ -1,24 +1,15 @@
...
 description: |
-  The best way to utilize the Custom Elements Language Server is to enable the Custom Elements Manifest (CEM) in your
-  project by installing a CEM generator like one provided by The Open WC Team.
-
-  Generating a CEM in watch mode will provide you with the best user experience. If your dependencies ship with a Custom
-  Elements Manifest, those will be utilized also.
+  Custom Elements Language Server provides useful language features for Web Components. Features include code actions, completons, diagnostics and more.
 homepage: https://github.com/Matsuuu/custom-elements-language-server
 licenses:
   - BSD-3-Clause
```
williamboman, line 7: The `languages` field is a required one so if there are no associated languages we'll need to:

```suggestion
  - BSD-3-Clause
languages: []
```

packages/custom-elements-languageserver/package.yaml
```diff
@@ -1,24 +1,15 @@
...
-  - HTML
-  - JavaScript
-  - TypeScript
-  - JSX
 categories:
   - LSP
 
 source:
-  id: pkg:npm/custom-elements-languageserver@0.0.8
+  id: pkg:npm/custom-elements-languageserver@1.0.1
```
williamboman, line 12: Hm I wonder why Renovate hasn't picked this up. I believe it may be because the current version is `<1`, will look into it.

williamboman (approved):
Thanks!

packages/custom-elements-languageserver/package.yaml
Matsuuu: I think it might be that I pushed them today too? Maybe it's just a bit slow. Or could be that it's a major version bump

williamboman: Ah yeah seems like [it was picked up ~15 hours ago](https://github.com/mason-org/mason-registry/pull/3393).

PR #3394, opened 2023-11-09
feat: add facility-language-server

bryankenote:
https://facilityapi.github.io/

Facilityapi is an open source interface definition language similar to protobuf. There are code generators for C#, Python, and Typescript. Currently they all implement http, but a facility service definition itself is protocol agnostic.


schemas/enums/language.json
```diff
@@ -51,6 +51,7 @@
         "Erg",
         "Erlang",
         "F#",
+        "FacilityServiceDefinition",
```
williamboman, line 54: We can include spaces here, I think it'd be more correct to use:

```suggestion
        "Facility Service Definition",
```

packages/facility-language-server/package.yaml
```diff
@@ -0,0 +1,28 @@
...
+
+source:
+  id: pkg:github/FacilityApi/FacilityLanguageServer@v2.1.0
+  asset:
+    - target: darwin_x64
+      file: Facility.LanguageServer.zip:libexec/
+    - target: darwin_arm64
+      file: Facility.LanguageServer.zip:libexec/
+    - target: win_x64
+      file: Facility.LanguageServer.zip:libexec/
```
williamboman, line 22: Should this not be `Facility.LanguageServer-win-x64.zip`? Also is there no Linux support?

ddunkin: `Facility.LanguageServer.zip` is platform neutral. `Facility.LanguageServer-win-x64.zip` contains an exe to make running on Windows easier for the Visual Studio extension.

packages/facility-language-server/package.yaml
```diff
@@ -0,0 +1,34 @@
...
+    - target: darwin_arm64
+      file: Facility.LanguageServer.zip:libexec/
+    - target: linux_x64
+      file: Facility.LanguageServer.zip:libexec/
+    - target: linux_arm64
+      file: Facility.LanguageServer.zip:libexec/
+    - target: win_x64
+      file: Facility.LanguageServer.zip:libexec/
+    - target: win_arm64
+      file: Facility.LanguageServer.zip:libexec/
```
williamboman, line 28: Apologies for the back and forth on this one, but we can simplify this to a single platform-agnostic expression:

```suggestion
  asset:
    file: Facility.LanguageServer.zip:libexec/
```

PR #3427, opened 2023-11-11
feat: add cypher-ls

meistermeier:
With vim file type support waiting for neovim to get picked up https://github.com/vim/vim/commit/8f0fe20ff1a13b468fdfaf85c434475f75c7a615
and
already existing support in nvim-lspconfig https://github.com/neovim/nvim-lspconfig/commit/cedc13473c7ff65bfc85cf3a9cfa65f7876a03b8
it would only make sense to me to also create the foundation (mason-registry) for this LSP to get added to mason-lspconfig in the next run.


packages/cypher-ls/package.yaml
```diff
@@ -0,0 +1,16 @@
...
+homepage: https://github.com/neo4j/cypher-language-support
+licenses:
+  - Apache-2.0
+languages:
+  - Cypher
+categories:
+  - LSP
+
+source:
+  id: pkg:npm/@neo4j-cypher/language-server@2.0.0-next.1
```
williamboman, line 13: The `@` sign will need to be [percent-encoded](https://en.wikipedia.org/wiki/Percent-encoding):
```suggestion
  id: pkg:npm/%40neo4j-cypher/language-server@2.0.0-next.1
```

williamboman (approved):
Thanks!

PR #3504, opened 2023-11-18
feat: add cds-lsp

stijnme:
Added SAP CDS language server for CAP projects.


stijnme:
Please let me know if (and how) any more tests on my side can be executed before pulling this into the Mason registry.

packages/cds-lsp/package.yaml
```diff
@@ -0,0 +1,18 @@
...
+licenses:
+  - proprietary
+languages:
+  - CDS
+categories:
+  - LSP
+  # - Formatter
+
+source:
+  id: pkg:npm/@sap/cds-lsp@7.4.0
```
williamboman, line 14: We'll have to encode the `@` sign, otherwise anything right of it will be parsed as the version component, per the PURL spec:

```suggestion
  id: pkg:npm/%40sap/cds-lsp@7.4.0
```

packages/cds-lsp/package.yaml
```diff
@@ -0,0 +1,18 @@
...
+
+bin:
+  cds-lsp: npm:cds-lsp --stdio
```
williamboman, line 17: These entries should only provide a target for the executable, so we can't include arguments here:

```suggestion
  cds-lsp: npm:cds-lsp
```

The command to actually run the LSP server should be configured in nvim-lspconfig, for example. Mason only makes the executable available.

williamboman (commented):
LGTM! I think we can uncomment the commented lines, unless they were commented for a reason?

stijnme:
> LGTM! I think we can uncomment the commented lines, unless they were commented for a reason?

Was it the good way to add also the formatter from the same package?

```
 categories:
   - LSP
   - Formatter

 source:
   id: pkg:npm/%40sap/cds-lsp@7.4.0

 bin:
   cds-lsp: npm:cds-lsp
   format-cds: npm:format-cds
```

I've removed the formatting part for now, to take it step by step.

williamboman:
> Was it the good way to add also the formatter from the same package?

Yeah it's a good way to include it since the underlying package contains those capabilities, let's include it again :+1:

stijnme:
> > Was it the good way to add also the formatter from the same package?
> 
> Yeah it's a good way to include it since the underlying package contains those capabilities, let's include it again 👍

Hi @williamboman,

First of all: thanks for your patience and quick responses!

I've added the formatting again, but I'm still not sure it will work.

Following is defined in `package.json`:
```
  "bin": {
    "cds-lsp": "./dist/main.js",
    "format-cds": "./scripts/formatCli.js"
  },
```
Meaning that `cds-lsp` is in fact `./dist/main.js`.

Therefor I should adjust the following?
```
bin:
  cds-lsp: npm:dist/main.js
  format-cds: npm:scripts/formatCli.js
```

Also in `package.json` there is a list of dependencies:
```
  "dependencies": {
    "@sap/cds-compiler": "4.4.0",
    "ignore": "5.2.4",
    "json-source-map": "0.6.1",
    "marked": "10.0.0",
    "tiny-typed-emitter": "2.1.0",
    "vscode-languageserver": "9.0.1",
    "vscode-languageserver-protocol": "3.17.5",
    "vscode-languageserver-textdocument": "1.0.11"
  }
```
These are indeed also installed when installing the package manually.

When looking at `typescript-language-server` should I configure these also as `extra`?
```
source:
  id: pkg:npm/typescript-language-server@4.1.2
  extra_packages:
    - typescript
```

With kind regards,
Stijn

williamboman:
So the `bin` entries in `package.json` is only relevant for npm. All these entries will end up inside `node_modules/.bin` when installing the package, which is the location that Mason targets (this is what `npm:cds-lsp` and `npm:format-cds` means).

> When looking at `typescript-language-server` should I configure these also as `extra`?

No these dependencies are automatically installed by npm. The reason `typescript-language-server` contains this is because they don't include `typescript` as a dependency, for whatever reason.

williamboman (approved):
LGTM!

PR #3513, opened 2023-11-19
feat: add v-analyzer

ttytm:


packages/v-analyzer/package.yaml
```diff
@@ -0,0 +1,32 @@
...
+homepage: https://github.com/v-analyzer/v-analyzer
+licenses:
+  - MIT
+languages:
+  - V
+categories:
+  - LSP
+
+source:
+  id: pkg:github/v-analyzer/v-analyzer@latest
```
williamboman, line 13: Using the `latest` release is problematic because it's an immutable release version that currently breaks the entire version management of the package. They seem to have provided an initial release, let's use that?
```suggestion
  id: pkg:github/v-analyzer/v-analyzer@v0.0.1-beta.1
```

packages/v-analyzer/package.yaml
```diff
@@ -0,0 +1,32 @@
...
+  asset:
+    - target: darwin_arm64
+      file: v-analyzer-darwin-arm64.zip
+      bin: v-analyzer
+    - target: darwin_x64
+      file: v-analyzer-darwin-x86_64.zip
+      bin: v-analyzer
+    - target: linux_x64
```
williamboman, line 21: I believe it's built for GNU systems:
```suggestion
    - target: linux_x64_gnu
```

packages/v-analyzer/package.yaml
ttytm: Thanks for the review @williamboman . I see so this is the literal tag, not what is available via https://github.com/v-analyzer/v-analyzer/releases/latest? 

If we do so, is it possible to update the version short term when there is a new release? I.e. very recently we just added optimisation for the linux release binary, this change isn't included in `@v0.0.1-beta.1`. Or would nightly be okay since it's a young project? Nightly releases are provided: https://github.com/v-analyzer/v-analyzer/releases/tag/latest

williamboman: > If we do so, is it possible to update the version short term when there is a new release? I.e. very recently we just added optimisation for the linux release binary, this change isn't included in `@v0.0.1-beta.1`.

Whenever there's a new release made Renovate will automatically pick it up and upgrade the version. While using the nightly version (`latest`) will technically be possible it will effectively render the version management of the package useless so it's not something we should do. If using the nightly build is the recommended way it might not be mature enough to be packaged in this registry yet.

If there are new changes that you'd like to get out it might make sense to cut a new release.

packages/v-analyzer/package.yaml
```diff
@@ -0,0 +1,32 @@
...
+      bin: v-analyzer
+    - target: linux_x64_gnu
+      file: v-analyzer-linux-x86_64.zip
+      bin: v-analyzer
+    - target: win_x64
+      file: v-analyzer-windows-x86_64.zip
+      bin: v-analyzer.exe
+
+schemas:
+  lsp: vscode:https://raw.github.com/v-analyzer/v-analyzer/v-analyzer/{{version}}/editors/code/package.json
```
williamboman, line 29:
```suggestion
  lsp: vscode:https://raw.githubusercontent.com/v-analyzer/v-analyzer/{{version}}/editors/code/package.json
```

packages/v-analyzer/package.yaml
```diff
@@ -0,0 +1,32 @@
...
+      bin: v-analyzer
+    - target: win_x64
+      file: v-analyzer-windows-x86_64.zip
+      bin: v-analyzer.exe
+
+schemas:
+  lsp: vscode:https://raw.githubusercontent.com/v-analyzer/v-analyzer/{{version}}/editors/code/package.json
+
+bin:
+  v-analzyer: "{{source.asset.file}}"
```
williamboman, line 32:
```suggestion
  v-analzyer: "{{source.asset.bin}}"
```

PR #3514, opened 2023-11-19
Add nilaway linter

ananthb:
[nilaway](https://github.com/uber-go/nilaway) is a static analysis tool that tries to catch nil panic errors in Go.

### DIFF
--- a/packages/nilaway/package.yaml
+++ b/packages/nilaway/package.yaml
@@ -14,7 +14,7 @@ categories:
   - Linter
 
 source:
-  id: pkg:golang/go.uber.org/nilaway/cmd/nilaway@a267567c6ffff900df0c3394d031ee70079ec8df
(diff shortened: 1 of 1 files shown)


packages/nilaway/package.yaml
```diff
@@ -0,0 +1,20 @@
...
+licenses:
+  - Apache-2.0
+
+languages:
+  - Go
+categories:
+  - Linter
+
+source:
+  id: pkg:golang/go.uber.org/nilaway/cmd/nilaway@a267567c6ffff900df0c3394d031ee70079ec8df
```
williamboman, line 17: According to the CI test suites this seems to already work but I believe Renovate will struggle with a commit SHA as the version specifier, so it'd be better to use the latest published version instead:
```suggestion
  id: pkg:golang/go.uber.org/nilaway@v0.0.0-20231117175943-a267567c6fff#cmd/nilaway
```

Also moved the subpath `cmd/nilaway` to the `#fragment` component of the PURL as `go.uber.org/nilaway` is the canonical package name

PR #3525, opened 2023-11-20
feat: add stimulus-language-server

tetzng:


packages/stimulus-language-server/package.yaml
```diff
@@ -0,0 +1,22 @@
...
+  - PHP
+  - Ruby
+categories:
+  - LSP
+
+source:
+  id: pkg:npm/stimulus-language-server@0.2.1
+
+schemas:
+  lsp: vscode:https://raw.githubusercontent.com/marcoroth/stimulus-lsp/main/package.json
```
williamboman, line 19: The configuration schema seems to only contain a single VSCode-only setting, but we might as well include it. To make sure the correct version of the schema is installed we can interpolate the `version` that is being installed:

```suggestion
  lsp: vscode:https://raw.githubusercontent.com/marcoroth/stimulus-lsp/v{{version}}/package.json
```

PR #3526, opened 2023-11-20
feat: add hydra-lsp

liepieshov:
Adding LSP for YAML configuration files based on Hydra python package for scientific projects https://hydra.cc/ 
https://github.com/Retsediv/hydra-lsp


williamboman (approved):
Thanks!

PR #3532, opened 2023-11-21
Version bump for tectonic

JuanCldCmt:
The renovate bot doesn't seem to pick up the update, and it's around 3 months since the update rolled out on June 15, 2023.

If maintenance is hard and barely anyone is using it (or noticing the outdated version), maybe we can just remove this package.

Also see https://github.com/williamboman/mason.nvim/pull/708


williamboman (approved):
Thanks and sorry for late review!

PR #3537, opened 2023-11-22
Add vscode-autohotkey2-lsp

kls1991:
Permission and license from https://github.com/thqby/vscode-autohotkey2-lsp/issues/437 (in Chinese)

The installing method seems a bit strange, but it's already the best I can get.

Tested with manully changing the huge registry.json, seems works.

Some lsp-servers comes with a json file under mason-schema, I didn't really figure out their usages, is the json file necessary?

This is my first pull request ever, hope I did things right!

### DIFF
--- a/packages/vscode-autohotkey2-lsp/package.yaml
+++ b/packages/vscode-autohotkey2-lsp/package.yaml
@@ -1,0 +1,20 @@
+---
+name: vscode-autohotkey2-lsp
+description: Autohotkey v2 Language Support using vscode-lsp.
+homepage: https://github.com/thqby/vscode-autohotkey2-lsp
+licenses:
+  - LGPL-3.0
+languages:
+  - AutoHotKey
+categories:
+  - LSP
+
+source:
+  # renovate:datasource=git-refs
+  id: pkg:github/thqby/vscode-autohotkey2-lsp@f358179042fd653a2c536ef24bb1b03898dfd5ca
+  build:
+    - target: win
+      run: |
+        cp ./tools/install.js .
+        rm * -Exclude install.js -Recurse -Force
+        node ./install.js


packages/vscode-autohotkey2-lsp/package.yaml
```diff
@@ -0,0 +1,20 @@
...
+  - LSP
+
+source:
+  # renovate:datasource=git-refs
+  id: pkg:github/thqby/vscode-autohotkey2-lsp@f358179042fd653a2c536ef24bb1b03898dfd5ca
+  build:
+    - target: win
+      run: |
+        cp ./tools/install.js .
+        rm * -Exclude install.js -Recurse -Force
```
williamboman, line 19: This is a nice idea to clean up unnecessary artifacts. However, this will be executed within a git repository so removing the files will not have any meaningful impact as they're tracked within `.git/` anyway.

packages/vscode-autohotkey2-lsp/package.yaml
```diff
@@ -0,0 +1,20 @@
...
+        node ./install.js
```
williamboman, line 20: I'm not a big fan of executing scripts like this for the following reasons:
- it may require additional system dependencies
- it may require a specific version of `node`
- there's no auditing or protection against the contents of it

It'd be much better if they were to publish it to `npm` instead. By being hosted on npm it's subject to removal and automatic auditing through services like https://socket.dev/.

williamboman (commented):
Thanks for the PR and apologies for late review. Would you mind exploring whether they'd be able to publish it as a package to npm instead?

williamboman:
> Some lsp-servers comes with a json file under mason-schema, I didn't really figure out their usages, is the json file necessary?

It's not strictly necessary. The contents of this JSON file is used to collect the configuration schema of the LSP server. It should only be added if there is such a configuration schema, which it doesn't seem to be in this case.

kls1991:
> Thanks for the PR and apologies for late review. Would you mind exploring whether they'd be able to publish it as a package to npm instead?

I asked the author to provide a npm package or github release, but he seems don't want to do that.
We talked in https://github.com/thqby/vscode-autohotkey2-lsp/issues/437

kls1991:
> Thanks for the PR and apologies for late review. Would you mind exploring whether they'd be able to publish it as a package to npm instead?

It's ok. Few people want to use neovim to write autohotkey......
Should I close this pull request? I don't know how to be more courteous on GitHub :p

fcying:
> > Thanks for the PR and apologies for late review. Would you mind exploring whether they'd be able to publish it as a package to npm instead?
> 
> It's ok. Few people want to use neovim to write autohotkey...... Should I close this pull request? I konw little about github convention :p

You can also register a local repository for ahk2-lsp to use

PR #3545, opened 2023-11-22
bump(latexindent): update to v.3.23.4

mathjiajia:


williamboman (approved):
Thanks! That capital V is breaking Renovate :(

PR #3546, opened 2023-11-22
feat: add typos-lsp

mehalter:


mehalter:
I believe I got the architectures specified correctly. Let me know if things like `linux_x64_gnu` should simply be `linux_x64`.

packages/typos-lsp/package.yaml
```diff
@@ -0,0 +1,41 @@
+---
+name: typos-lsp
+description: Source code spell checker
+homepage: https://github.com/tekumara/typos-vscode
+licenses:
+  - MIT
+  - Apache-2.0
```
williamboman, line 7: I can't see Apache-2.0 being referred to as their license anywhere

williamboman (approved):
LGTM! Just one question re: licenses

packages/typos-lsp/package.yaml
mehalter: Oh yeah that's totally a mistake, copy/paste error. My bad! Resolved in the latest commit

PR #3559, opened 2023-11-23
add darwin arm64 for `typos`

chrisgrieser:
So `typos` [does not have a dedicated built for arm64](https://github.com/crate-ci/typos/issues/828), but it does work fine on my arm64 machine. 

So I suppose it wouldn't hurt to enable installing it via mason, even if it is not optimized? (Only added support for mac in this PR, since I do not have a linux/windows device and therefore cannot confirm that `typos` works there as well.)

### DIFF
--- a/packages/typos/package.yaml
+++ b/packages/typos/package.yaml
@@ -15,6 +15,10 @@ source:
     - target: darwin_x64
       file: typos-{{ version }}-x86_64-apple-darwin.tar.gz
       bin: typos
+    # No arm64 build yet, but does work fine on M1 Mac. See https://github.com/crate-ci/typos/issues/828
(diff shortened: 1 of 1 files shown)


williamboman (approved):
Thanks! The `target` field allows an array of values, so I usually reduce it to:

```
    - target: [darwin_x64, darwin_arm64]
      file: typos-{{ version }}-x86_64-apple-darwin.tar.gz
      bin: typos
```

As for compatibility, most Intel (x64) binaries run fine on the Apple chips (arm64) because of [Rosetta translation](https://en.wikipedia.org/wiki/Rosetta_(software)). There's overhead of course but for most use cases it's negligible.

PR #3578, opened 2023-11-26
bump(beancount-language-server): update to v1.3.3

Quanuanc:


williamboman:
Thanks for the PR! I was unable to merge this before because it didn't recognize 1.3.3 as a valid version. Renovate has now updated it to 1.3.4, so we can close this :)

PR #3603, opened 2023-11-28
feat: add hlint

Conarius:


Conarius:
I think there needs to be a runner/option that supports installation through cabal and stack as they are commonly used for Haskell dev tools.
Have looked into it and if there is any work to be done I will gladly help out

williamboman:
> I think there needs to be a runner/option that supports installation through cabal and stack as they are commonly used for Haskell dev tools. Have looked into it and if there is any work to be done I will gladly help out

Hey! Yeah `pkg:hackage` is not a supported source type in Mason yet. For it to be added it needs to support installation of packages in such a way that:
1) installation artifacts are entirely isolated in a given directory on the filesystem (excluding cache, log files etc.)
2) resulting artifacts are executable in a universal fashion (i.e. no implicit runtime dependencies)

I don't believe Cabal is capable of this as it heavily utilizes `~/.local/` to permanently store build artifacts.

Conarius:
So I guess the other option would be via GitHub releases, as they're luckily provided on the GitHub page

Conarius:
@williamboman
I've updated the package definition and now it works. It uses Github release assets.

PR #3607, opened 2023-11-28
bump(jq-lsp): update to v0.0.0-20231016080857-b4707e7776a4

mattmeyers:
Hello. It looks like this LSP version has never been updated since originally being added. Since then there have been some bug fixes in the LSP.

### DIFF
--- a/packages/jq-lsp/package.yaml
+++ b/packages/jq-lsp/package.yaml
@@ -12,7 +12,7 @@ categories:
   - LSP
 
 source:
-  id: pkg:golang/github.com/wader/jq-lsp@v0.0.0-20221220100930-4aadc9e4bf54
+  id: pkg:golang/github.com/wader/jq-lsp@v0.0.0-20231016080857-b4707e7776a4
 
 bin:
   jq-lsp: golang:jq-lsp


williamboman:
Thanks! It should've been picked up by Renovate. After reviewing the logs it seems like it ends up using the semver versioning strategy which fails because it's not semver compliant. I'd have assumed that Renovate would use Go's versioning strategy. Let's see if this gets picked up by Renovate via https://github.com/mason-org/mason-registry/pull/3648 (has to be merged first), if not we'll unfortunately have to manually manage the version for the time being.

williamboman:
Never mind, it seems like the old version has been pulled and `v0.0.0-20231016080857-b4707e7776a4` is the only one. I'd like to believe this is an issue with how they version it, it'd be nice if it was actual versions.

mattmeyers:
Hey @williamboman, I’m not quite sure why CI is failing here. Anything I need to do to get it passing?

williamboman:
> Hey @williamboman, I’m not quite sure why CI is failing here. Anything I need to do to get it passing?

It's failing because the specified version is not available in the Golang pkg registry API. I'm not entirely familiar with how they allow versioning of packages but I assume in this case the versions are sourced upstream from the git repo.

> Hey, now there is https://github.com/wader/jq-lsp/releases/tag/v0.1.0

Sweet! That version went through automatically without any issues, https://github.com/mason-org/mason-registry/pull/3700, so we can close this PR

PR #3646, opened 2023-12-01
chore(pkgbuild-language-server): add deprecation

williamboman:
The package seems to have completely disappeared from the internet.


williamboman:
The old GitHub repo redirects to https://github.com/Freed-Wu/tree-sitter-lsp 🤷

PR #3662, opened 2023-12-03
Added support for hdl_checker lsp

fred441a:
added [HDL_checker](https://github.com/suoto/hdl_checker) to mason registry.

goes together with [this](https://github.com/williamboman/mason-lspconfig.nvim/pull/336) pull request on mason-lspconfig


williamboman (approved):
Thanks!

PR #3683, opened 2023-12-05
bump(haskell-language-server): update to version 2.5.0.0

mystilleef:
I updated the supported prebuilt GHC binaries.


williamboman (approved):
Thanks!

PR #3707, opened 2023-12-07
feat: add termux-language-server

Relms12345:
as per #3646, pkgbuild-language-server was wiped from the internet. The successor to that is termux-language-server. It adds several other file formats other then a PKGBUILD. This PR adds that LSP and I recommend deprecating or removing pkgbuild-language-server.

### DIFF
--- a/packages/termux-language-server/package.yaml
+++ b/packages/termux-language-server/package.yaml
@@ -1,0 +1,16 @@
+---
+name: termux-language-server
+description: A language server for some specific bash scripts.
+homepage: https://termux-language-server.readthedocs.io/en/latest/
+licenses:
+  - GPL-3.0
+languages:
+  - Bash
+categories:
+  - LSP
+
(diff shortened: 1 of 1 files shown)


williamboman (commented):
Hey, thanks! I'm however not too keen on including projects that simply disappear from the internet (such as pkgbuild). How mature is this package? Are there any indications that suggests it will not also be wiped?

Conarius:
From what I've seen so far is that it is put under the termux community and it receives a lot of commits from one dude, who seems to be the maintainer for it.
Featurewise it is almost complete, only code actions are completely missing. For PKGBUILD itself it has everything you need from a LSP server.

williamboman:
I see! I've gone ahead and [asked the maintainer](https://github.com/termux/termux-language-server/discussions/10) if they have any input regarding adding it here yet.

PR #3724, opened 2023-12-08
feat: add oelint-adv

antznin:
Linter for [bitbake](https://en.wikipedia.org/wiki/BitBake) recipes.

### DIFF
--- a/packages/oelint-adv/package.yaml
+++ b/packages/oelint-adv/package.yaml
@@ -1,0 +1,15 @@
+---
+name: oelint-adv
+description: Linter for bitbake recipes.
+homepage: https://github.com/priv-kweihmann/oelint-adv
+licenses:
+  - BSD-2-Clause
+languages: []
+categories:
+  - Linter
+
+source:
+  id: pkg:pypi/oelint-adv@3.26.3
+
+bin:
+  oelint-adv: pypi:oelint-adv


packages/oelint-adv/package.yaml
```diff
@@ -0,0 +1,15 @@
+---
+name: oelint-adv
+description: Linter for bitbake recipes.
+homepage: https://github.com/priv-kweihmann/oelint-adv
+licenses:
+  - BSD-2-Clause
+languages: []
```
williamboman, line 7: I have no idea what bitbake is, after some searching I wonder if it'd make sense to add the following languages?

- YOCTO
- OpenEmbedded

williamboman (approved):
Thanks!

packages/oelint-adv/package.yaml
antznin: I guess the language to use would be Bitbake. Bitbake language support is native in neovim and the corresponding filetype is `bitbake`.

I didn't include it here at first because I didn't find it in  `schemas/enums/language.json` (and I thought it had to be in this list). Not knowing what the role of `languages:` is, I chose to let it empty. If this field is purely for description purposes, I would add these languages:

* Bitbake
* Yocto
* OpenEmbedded

If that looks good to you I can make another PR.

williamboman: > I didn't include it here at first because I didn't find it in `schemas/enums/language.json` (and I thought it had to be in this list). Not knowing what the role of `languages:` is, I chose to let it empty.

Yeah sometimes a language have simply not been added before, in which case it's perfectly fine to just include it in the schema.

> If this field is purely for description purposes, I would add these languages:

Yep it's for now entirely for descriptive purposes, to aid discovery etc. Since the language list is well defined in the schema it might also be used to translate to vim/neovim filetypes in the future, but there's no plans to do so yet.

> If that looks good to you I can make another PR.

Looks good to me :+1:, thanks!

PR #3740, opened 2023-12-09
feat(slint): add build config for mac

lynndylanhurley:


williamboman (commented):
Thanks! I'm not sure adding a dependency on Rust necessarily improves platform support. This of course always need to be evaluated in a context - what relation does Slint have with Rust? Would the average user have Rust installed?

ogoffart:
Rust doesn't need to be installed to use Slint, it is only used to compile it. But Slint can be used from different programming language, and we provide binaries so there is no need to have the rust tool chain.

Would it be possible to use the binary for platforms for which they are available, and build from source otherwise?

lynndylanhurley:
> Would it be possible to use the binary for platforms for which they are available, and build from source otherwise?

That's what I had initially. I'll revert.

lynndylanhurley:
@ogoffart @williamboman this PR has been updated to only use cargo for mac. can you please review?

williamboman:
> @ogoffart @williamboman this PR has been updated to only use cargo for mac. can you please review?

Hey! So unfortunately the package spec doesn't allow for multiple different source types (asset file vs build), so this won't pass schema validation nor be parseable by the mason.nvim client. This is by design to avoid convoluted definitions and ensure some sense of reproducibility across systems (it also makes maintenance easier). There really only are two options:

1) Change the entire package specification here to use `cargo` as the source (meaning cargo will be required for all users for installation). I don't believe this would be a good option seeing as there probably is a rather small overlap between Slint users and having Rust + Cargo installed on the system.
2) Add macOS release binaries upstream

lynndylanhurley:
Thanks @williamboman. 

@ogoffart are there any plans to add the macOS release binaries? if not then shouldn't we switch to `cargo` so there's at least some mac support?

ogoffart:
This is unfortunate that it is not possible to build from source for platform that don't have binaries.  We could add the macOS binaries, but that still wouldn't work for eg, Linux on Arm or RiscV or other more exotic OS or Platform.

Regarding macOS specifically, I opened an issue: https://github.com/slint-ui/slint/issues/4240

We actually already build the macOS binaries, but they are not part of the release because they don't work. I'm not using macOS, so i'm not familiar with the exact problem.  (Maybe @tronical can elaborate?)

tronical:
Yes, for the lsp we create a universal Mac binary. We need to sign it, still.

For this registry, what’s the best format of making the binary available? Our .zip file would contain additional file such as attributions.

packages/slint-lsp/package.yaml
```diff
@@ -19,6 +19,10 @@ source:
     - target: win_x64
       file: slint-lsp-windows.zip
       bin: slint-lsp/slint-lsp.exe
+    - target: [darwin_x86, darwin_arm64]
+      build:
+        run: cargo install slint-lsp
+      bin: slint-lsp
```
tronical, line 25: Since we were already building binaries for macOS, I took the liberty of adding them the 1.3.2 release at https://github.com/slint-ui/slint/releases/tag/v1.3.2 . Those `-macos.tar.gz` archives were produced in the same build as the linux/windows once. They are not code signed though, but from what I can tell other macOS binaries in the mason registry don't do that either (for example tree-sitter).

Could you try the suggestion below?

```suggestion
    - target: [darwin_x86, darwin_arm64]
      file: slint-lsp-macos.tar.gz
      bin: slint-lsp/slint-lsp
 ```

**Comment by lynndylanhurley** [comment]
@tronical I just commited that change. any other blockers to getting this merged?

**Comment by tronical** [comment]
Sorry, I’m not an approver in this project :(. Do you know how to test this change?

File: packages/slint-lsp/package.yaml [hunk 2]
```diff
@@ -19,6 +19,9 @@ source:
     - target: win_x64
       file: slint-lsp-windows.zip
       bin: slint-lsp/slint-lsp.exe
+    - target: [darwin_x86, darwin_arm64]
```
williamboman, line 22:
```suggestion
    - target: [darwin_x64, darwin_arm64]
```

williamboman (approved):
Thanks! I very recently added support for OpenVSX sources, so in the case publishing via GitHub turns out to be problematic we can use the [OpenVSX distribution](https://open-vsx.org/extension/Slint/slint)

tronical:
Thank you for reviewing and merging. I was able to verify that installing `slint_lsp` from mason works now also on macOS out of the box. Yay :)

PR #3767, opened 2023-12-12
feat: Add htmlhint

dante0624:
### Use Case
I am aware that the mason registry already has html linters such as markuplint, however I have found htmlhint to be the easiest to use. I think it would integrate really nicely with plugins like nvim-lint, as the binary file seems to work right out of the box.


williamboman (approved):
Thanks!

PR #3774, opened 2023-12-12
bump(staticcheck): update to v0.4.6

radimsuckr:
Currently used version is several months old and the latest is `v0.4.6`.

Package docs for reference: https://pkg.go.dev/honnef.co/go/tools@v0.4.6/cmd/staticcheck?tab=versions

### DIFF
--- a/packages/staticcheck/package.yaml
+++ b/packages/staticcheck/package.yaml
@@ -10,7 +10,7 @@ categories:
   - Linter
 
 source:
-  id: pkg:golang/honnef.co/go/tools/cmd/staticcheck@v0.4.6
+  id: pkg:golang/honnef.co/go/tools@v0.4.6#cmd/staticcheck
 
 bin:
   staticcheck: golang:staticcheck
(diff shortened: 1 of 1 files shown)


williamboman (approved):
Thanks! Renovate probably struggled with reconciling the version changes due to the current version being in a different format.

packages/staticcheck/package.yaml
```diff
@@ -10,7 +10,7 @@ categories:
   - Linter
 
 source:
-  id: pkg:golang/honnef.co/go/tools/cmd/staticcheck@2023.1.3
+  id: pkg:golang/honnef.co/go/tools/cmd/staticcheck@v0.4.6
```
williamboman, line 13: Actually, we might as well change the PURL so that it targets the canonical package name (this might be why Renovate didn't manage to update):

```suggestion
  id: pkg:golang/honnef.co/go/tools@v0.4.6#cmd/staticcheck
```

radimsuckr:
To be honest I don't know much about Go packages nor renovate bot. I just didn't like the errors of older staticcheck versions when doing Advent of Code and decided to try and update the version in registry. :smile:

PR #3809, opened 2023-12-15
chore(oelint-adv): update languages

antznin:
Update languages following discussion on #3724.

### DIFF
--- a/packages/oelint-adv/package.yaml
+++ b/packages/oelint-adv/package.yaml
@@ -6,8 +6,6 @@ licenses:
   - BSD-2-Clause
 languages:
   - Bitbake
-  - Yocto
-  - OpenEmbedded
 categories:
   - Linter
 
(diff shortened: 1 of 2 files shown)


antznin:
@williamboman the pipeline is failing. I'm guessing this is because of the missing languages in `languages.json`?

I understood from our discussion on #3274 that it wasn't required. If it is then I would only add Bitbake, as Yocto and OpenEmbedded are not languages per se but rather the organizations maintaining the Bitbake language/tools.

williamboman:
> I understood from our discussion on #3274 that it wasn't required. If it is then I would only add Bitbake, as Yocto and OpenEmbedded are not languages per se but rather the organizations maintaining the Bitbake language/tools.

It is indeed required, I pushed a commit that adds `Bitbake` to schema and removed the other entries from languages :+1:

williamboman (approved):
Thanks!

PR #3855, opened 2023-12-19
feat: add twig-language-server

huesersohn:


packages/twig-language-server/package.yaml
```diff
@@ -0,0 +1,22 @@
...
+categories:
+  - LSP
+  - Linter
+
+source:
+  # renovate:datasource=github-tags
+  id: pkg:generic/moetelo/twiggy@v0.6.2
+  download:
+    files:
+      twiggy.vsix: https://moetelo.gallery.vsassets.io/_apis/public/gallery/publisher/moetelo/extension/twiggy/{{ version | strip_prefix "v" }}/assetbyname/Microsoft.VisualStudio.Services.VSIXPackage
```
williamboman, line 19: This does not seem like a stable URL nor is it allowed to use directly per VSCode marketplace terms of service.

The best option would be if they were to publish their https://github.com/moetelo/twiggy/tree/master/packages/language-server package to npm. It seems like there already is a `twig-language-server` on npm, provided by the upstream fork.

huesersohn: Alright, I've asked over there: https://github.com/moetelo/twiggy/issues/5

Should I keep this PR open and update it once there is another way to install it? I thought about building it from source but was not sure how to go about it with the way the repository is structured. Would that also be an option?

williamboman: > Should I keep this PR open and update it once there is another way to install it? I thought about building it from source but was not sure how to go about it with the way the repository is structured. Would that also be an option?

Building from source is a last resort when all other options have been exhausted. I'd like to shift responsibility to package maintainers themselves whenever it makes sense as it'd benefit everyone. Let's keep this PR open for the time being, it should be in their interest to publish to npm.

huesersohn: @williamboman The package is now published as twiggy-language-server. The maintainer asked to rename the package in this PR. Should I just update here or create a new PR?

williamboman: Cool! Let's update here :)

SushyDev: any updates on this?

huesersohn: @SushyDev I'm on it, just had no time the last weeks. And now I just noticed that the package contains neither source code nor a built version. I've updated my issue at the source. Sorry for the wait!

huesersohn:
@williamboman Ok, I've updated the PR with the requested changes.

williamboman (approved):
Thanks! It'd be nice if they also provided a `bin` entry in their `package.json`, pointing to the `server.js` so that it's recognized as an executable by npm itself as well!

SushyDev:
Installing the plugin through mason went successful but opening up a twig file and running `:LspInfo` does not show twiggy running and it doesn't seem to be running at all in the buffer either.

PR #3867, opened 2023-12-20
bump(sonarlint-language-server): update to v4.2.2

thahnen:
(Copied from the last PR because it is still the same)

Once again, not semver-compliant 😞 At last after Christmas I'll find the time to create an automation on our side to find the newest version and create a PR ^^


williamboman (approved):
Thanks!

PR #3896, opened 2023-12-21
feat: add checkmake

Conarius:
checkmake is a linter for makefiles written in go


williamboman (approved):
Thanks! I changed to pull the binary from their GitHub releases as this package does seem to be hosted on https://pkg.go.dev/, which is a requirement for Go packages via Mason.

PR #3899, opened 2023-12-21
feat: add cortex-debug

jedrzejboczar:
Adds https://github.com/Marus/cortex-debug for use with https://github.com/jedrzejboczar/nvim-dap-cortex-debug.


williamboman (commented):
Thanks! Do you know if these `.vsix` assets are available for stable releases as well? I believe they only publish them for pre-releases to make it possible to install it outside the VSCode marketplace.

It seems like it's distributed [via OpenVSX](https://open-vsx.org/extension/marus25/cortex-debug), which would be a better option if we can't rely on GitHub releases (I'm currently working on adding OpenVSX support in the `mason.nvim` client)

jedrzejboczar:
Unfortunately it looks like they don't publish stable releases on GitHub. If mason.nvim gets support for OpenVSX then I think this would be the best option.

williamboman:
I changed to use `openvsx` as source. I also removed the `share` entry because it linked too much - instead I think we can simply provide an executable (`cortex-debug`) for easy access to the debug adapter, for example via nvim-dap:

```lua
dap.adapters.cortex = {
  type = 'executable',
  command = 'cortex-debug',
  args = {}
}
```

What do you think? The VSCode extension seems to do a lot of auxiliary things, almost 8k LOC itself. How usable is this without access to VSCode client-specific features?

jedrzejboczar:
Hi, downloading with `openvsx` works well, but I'm not sure if providing just the `cortex-debug` executable is enough. 

The cortex-debug extension has the VSCode client part (emulated by nvim-dap-cortex-debug) and server part. The client must provide [launch_config.extensionPath](https://github.com/jedrzejboczar/nvim-dap-cortex-debug/blob/master/lua/dap-cortex-debug/adapter.lua#L180) to the server, and the server then uses this value in multiple places, e.g. https://github.com/Marus/cortex-debug/blob/master/src/gdb.ts#L935.

I could use something like `vim.env.MASON .. 'packages/cortex-debug'` instead of `vim.env.MASON .. '/share/cortex-debug'`, but is this ok to refer to things under `packages/`? I thought that `packages/` is somewhat private to mason and plugins should use the things in "installed" directories.

Also, providing `cortex-debug` as executable it is a bit inconvenient from my perspective, as I need special handling here: https://github.com/jedrzejboczar/nvim-dap-cortex-debug/blob/master/lua/dap-cortex-debug/adapter.lua#L187-L188 (instead of always just starting `command = node, args = debugadapter.js`). But this is a minor issue that I can easily solve. 

Another minor issue is that `cortex-debug` is a wrapper around `node`, so `node` must be still available on the system, but now users cannot provide another `node_path` in `require('dap-cortex-debug').setup()`.

williamboman:
Ah I see! Let's revert back to the share approach then. This way you can expand the path like so: `vim.fn.expand("$MASON/share/cortex-debug")`.

> I could use something like `vim.env.MASON .. 'packages/cortex-debug'` instead of `vim.env.MASON .. '/share/cortex-debug'`, but is this ok to refer to things under `packages/`? I thought that `packages/` is somewhat private to mason and plugins should use the things in "installed" directories.

Yep, that's correct!

williamboman:
I believe this should be good to go now. I'll go ahead and merge so that we can finally merge this, we can adjust things in later PRs if needed.

williamboman:
CI has gotten stuck 🙈, I'll merge this through another PR (I'll keep attribution to you)

jedrzejboczar:
Thanks!

Issue #3888, opened 2023-12-21
Add `mdformat-gfm` for Github Flavored Markdown support


tnthi115:
> @williamboman @anand-bala Is it possible to add [`mdformat-gfm`](https://pypi.org/project/mdformat-gfm/) and other options listed [here](https://github.com/executablebooks/mdformat#installing)? They seem to each have their own `mdformat` executable.

_Originally posted by @tnthi115 in https://github.com/mason-org/mason-registry/issues/2132#issuecomment-1865412278_

tnthi115:
When trying to push a new feature branch, I get the following error:

```sh
$ git push --set-upstream origin feat/add-mdformat-gfm
fatal: Could not read from remote repository.

Please make sure you have the correct access rights
and the repository exists.
```

PR #3900, opened 2023-12-21
feat: add mdformat-gfm (#3888)

tnthi115:
`mdformat-gfm` is an `mdformat` plugin that changes the target specification to GitHub Flavored Markdown (GFM).

Resolves: #3888

### DIFF
--- a/packages/mdformat-gfm/package.yaml
+++ b/packages/mdformat-gfm/package.yaml
@@ -1,0 +1,16 @@
+---
+name: mdformat-gfm
+description: Mdformat plugin for GitHub Flavored Markdown compatibility.
+homepage: https://github.com/hukkin/mdformat-gfm/
+licenses:
+  - MIT
+languages:
+  - Markdown
+categories:
+  - Formatter
+
+source:
+  id: pkg:pypi/mdformat-gfm@0.3.5
+
+bin:
+  mdformat: pypi:mdformat


williamboman:
Hello! Mason installs pypi packages in their own environments. As I understand it this is a plugin that needs to be installed in the same environment as `mdformat`? In this case this can't be added as a separate package but instead as a "sub-package" of mdformat, which is currently not yet supported.

UtkarshKunwar:
Since there are multiple plugins supported by `mdformat` as mentioned in their [documentation](https://mdformat.readthedocs.io/en/stable/users/plugins.html#id1) and they require being installed together.

I have a setup which contains `mdformat`, `mdformat-gfm`, `mdformat-toc`, and `mdformat-myst` (for not messing up math equations when the formatter runs). For now, I'll create a fork and try to have my own custom registry but it would be great if having plugins (multiple tools in the same virtualenv) gets supported upstream one day! :)

For a quick hack is there a way to get the path of the virtualenv where `mdformat` gets installed right now by Mason? That way maybe we can manually install the additional plugins we require for our configs.

williamboman:
> For a quick hack is there a way to get the path of the virtualenv where `mdformat` gets installed right now by Mason? That way maybe we can manually install the additional plugins we require for our configs.

For automated installations you can do something like below (haven't tested but you get the idea)

```lua
local registry = require("mason-registry")

registry.refresh(function ()
  local mdformat = registry.get_package("mdformat")
  mdformat:on("install:success", function ()
    local python = mdformat:get_install_path() .. "/venv/bin/python"
    vim.fn.jobstart({ python, "-m", "pip", "install", "mdformat-gfm" })
  end)
end)
```

This of course doesn't contain any error handling and won't work on Windows.

UtkarshKunwar:
Thanks @williamboman! Finally, I ended up doing it this way by putting this after the `mason.setup()` call.

```lua
local registry_status_ok, mason_registry = pcall(require, "mason-registry")
if not registry_status_ok then
    return
end

mason_registry.refresh(function()
    local mdformat = mason_registry.get_package("mdformat")
    local mdformat_extensions = {
        "mdformat-gfm",
        "mdformat-toc",
        "mdformat-myst",
    }
    mdformat:on("install:success", function()
        -- Create the installation command.
        vim.notify("Installing mdformat extensions.")
        local extensions = table.concat(mdformat_extensions, " ")
        local python = mdformat:get_install_path() .. "/venv/bin/python"
        local pip_cmd =
            string.format("%s -m pip install %s", python, extensions)

        -- vim.fn.jobstart doesn't work in callback so use popen instead.
        local handle = io.popen(pip_cmd)
        if not handle then
            vim.notify(
                'Could not install "mdformat extensions".',
                vim.log.levels.ERROR
            )
            return
        end
        local _ = handle:read("*a")
        handle:close()

        vim.notify('"mdformat extensions" were successfully installed.')
    end)
end)
```

PR #3949, opened 2023-12-28
Update elixir-ls to v0.18.1 with update to debugger

sax:
This is an attempt to reconcile the issues raised in #3910.

I have been unable to configure my mason installation in a way that uses my forked registry, so I'm unclear on whether any other change needs to be made to allow this to work.


williamboman (approved):
Thanks!

PR #3988, opened 2024-01-02
feat: add lemonade clipboard utility

WhoIsSethDaniel:


WhoIsSethDaniel:
There are packages for this utility but the current maintainer doesn't seem to make them often, if at all.  I attempted to have this build the most recent lemonade (which isn't very recent, but more recent than the last packages were made).

WhoIsSethDaniel:
@williamboman I'm not certain how to satisfy the package tests. I don't want mason to download packages (see above comment for why).

williamboman (commented):
Hello! I fail to see how this is related to Neovim? Seems out of scope for the core registry

WhoIsSethDaniel:
> Hello! I fail to see how this is related to Neovim? Seems out of scope for the core registry

lemonade is supported by Neovim as a clipboard tool (see `:h clipboard-tool`). In some environments, including my own, it's the only viable way to have two-way clipboard sharing between a remote nvim and a local nvim.  So having it installed is required to have Neovim work as expected.

williamboman:
Ah, I see. I still think it's somewhat in a gray area (evident by the empty languages and categories), but I think there's a strong enough use case for it so let's add it :)

The `golang` source is currently restricted to packages published to https://pkg.go.dev/ (and their underlying versions as provided by the pkg.go.dev API). They do seem to have cut a release in their latest commit 2 years ago (1.1.2) - let's use that instead (I'll update it right away and merge).

PR #4009, opened 2024-01-04
feat: idris2-lsp

### LINKED ISSUE #3934: idris2-lsp
Add idris2-lsp: https://github.com/idris-community/idris2-lsp. Could this interest people?

foxyseta:
Closes #3934.

### DIFF
--- a/packages/idris2-lsp/package.yaml
+++ b/packages/idris2-lsp/package.yaml
@@ -1,0 +1,25 @@
+---
+name: idris2-lsp
+description: Language Server for Idris2
+homepage: https://github.com/idris-community/idris2-lsp?tab=readme-ov-file
+licenses:
+  - BSD-3-Clause
+languages:
+  - Idris2
+categories:
+  - LSP
+
+source:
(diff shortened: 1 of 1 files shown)


foxyseta:
Bump

williamboman (commented):
Hello and thanks for the PR! I'm not too familiar with pack, but this installs the "app" globally, no? I've tried to find docs and reviewed the source for pack but was not able to find much information regarding where it installs things and whether it's configurable.

A hard requirement for a package to be added to Mason is that artifacts are only ever installed in Mason's package directory (defaults to current working directory for these build scripts). Things like cache and log files may however be written elsewhere.

foxyseta:
I agree. I guess there would be https://github.com/idris-community/idris2-lsp?tab=readme-ov-file#manual-installation, but it is probably not worth it. Do you think the feature we are looking for could be the missing "Support for custom build directories" at https://github.com/stefan-hoeck/idris2-pack#stuff-still-missing?

PR #4045, opened 2024-01-07
Decouple Erlang Language Server and Debugger

robertoaloi:
The debugger has been extracted from the Erlang LS repository, so it can be used in combination with other language servers (e.g. [ELP](https://github.com/mason-org/mason-registry/tree/main/packages/elp)). This PR (completely untested since I don't use Vim myself) adds a new package for the DAP debugger and removes the DAP parts from Erlang LS.

### DIFF
--- a/packages/erlang-debugger/package.yaml
+++ b/packages/erlang-debugger/package.yaml
@@ -1,0 +1,26 @@
+---
+name: erlang-debugger
+description: |
+  Step by step debugger for Erlang, powered by the Erlang/OTP interpreter and following the DAP protocol.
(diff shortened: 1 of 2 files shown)


robertoaloi:
Not sure what's the best way to notify users that a new package is required for the debugger. Another option could be to have it as a dependency (if those are supported by the registry).

fridayy:
>  Another option could be to have it as a dependency (if those are supported by the registry).

Afaik not supported. At least I could not find any reference in the other available lsp packages that have lsp and dap separated.

williamboman (approved):
Great, thanks!

PR #4114, opened 2024-01-12
feat: sqls repo was moved. re-point it in mason-registry

jason-h-35:
The original sqls repo was archived, with future development moved into the github.com/sqls-server GitHub org. 

This PR:
* Updates mason-registry to use https://github.com/sqls-server/sqls instead of https://github.com/lighttiger2505/sqls.
* Version bumps `sqls` from `v0.2.22` to `v0.2.27`.


williamboman (approved):
Thanks!

PR #4134, opened 2024-01-14
feat: Add cppcheck

haarisr:


haarisr:
@williamboman bumping this

williamboman:
Hey and thanks for the PR (and sorry for very late review)! I've been holding off with this PR because custom build scripts are being deprecated because of difficulty in maintaining them over time and their common reliance on system dependencies that Mason cannot detect. You can always host your own registry if you'd like to manage it through Mason, see https://github.com/williamboman/mason-registry-playground/ for a starting point

PR #4311, opened 2024-01-27
fix(ltex): use platform-specific releases for win/mac

chrisgrieser:
The main intent is to remove the external dependency on a JRE, thus fixing: https://github.com/williamboman/mason.nvim/issues/1531 & https://github.com/valentjn/ltex-ls/issues/187

Implements #2837, but also for mac. Linux still uses the platform-independent release, [to ensure compatibility with arm64](https://github.com/mason-org/mason-registry/pull/2837#pullrequestreview-1686482498).

### DIFF
--- a/packages/ltex-ls/package.yaml
+++ b/packages/ltex-ls/package.yaml
@@ -15,16 +15,25 @@ categories:
 source:
   id: pkg:github/valentjn/ltex-ls@16.0.0
   asset:
-    - target: unix
-      file: ltex-ls-{{version}}.tar.gz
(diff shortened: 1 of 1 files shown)


williamboman (approved):
This seems reasonable! It seems like it'll only end up using the bundled Java runtime if `$JAVA_HOME` is unset. This is probably an issue best directed upstream (maybe the release assets with a bundled runtime shouldn't respect `$JAVA_HOME`)

chrisgrieser:
indeed, with the bundled JRE, there is no need anymore to set `JAVA_HOME`. Yet another problem solved.

PR #4402, opened 2024-02-02
feat: add aiken

Conarius:


Conarius:
@williamboman
everything works now, had to rebase many branches xD

williamboman (approved):
Thanks! I noticed they provide prebuilt binaries which they themselves use so let's use that instead

PR #4471, opened 2024-02-07
fix(perl-debug-adapter): name change from vscode-perl-debug

WhoIsSethDaniel:


williamboman (approved):
Thanks! I saw that they tag releases so I took the opportunity to change version tracking as well

Issue #4443, opened 2024-02-06
Add glsl-analyzer


vypxl:
[GLSL Analyzer](https://github.com/nolanderc/glsl_analyzer) seems to be the best languageserver for glsl at this time. Would be nice to have it. They provide binary releases via Github.

PR #4578, opened 2024-02-16
feat: add glsl_analyzer

mehalter:
This resolves https://github.com/williamboman/mason.nvim/issues/625

### DIFF
--- a/packages/glsl_analyzer/package.yaml
+++ b/packages/glsl_analyzer/package.yaml
@@ -1,0 +1,39 @@
+---
+name: glsl_analyzer
+description: |
+  Language server for GLSL (autocomplete, goto-definition, formatter, and more)
+homepage: https://github.com/nolanderc/glsl_analyzer
+licenses:
+  - GPL-3.0
+languages:
+  - OpenGL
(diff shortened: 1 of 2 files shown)


Conarius:
Closes #4443

packages/glsl_analyzer/package.yaml
```diff
@@ -0,0 +1,39 @@
...
+    - target: win_x64
+      file: x86_64-windows.zip
+      bin: bin/glsl_analyzer.exe
+    - target: win_arm64
+      file: aarch64-windows.zip
+      bin: bin/glsl_analyzer.exe
+
+schemas:
+  lsp: vscode:https://raw.githubusercontent.com/nolanderc/glsl_analyzer-vscode/main/package.json
+
```
williamboman, line 37: This schema currently only contains one configuration item which is only used by the VSCode client extension: `glsl-analyzer.path` (sets the path to the executable). I think it'd mostly add noise and confusion so let's exclude the schema for now:

```suggestion
```

williamboman (approved):
Thanks!

PR #4591, opened 2024-02-17
feat: add spyglassmc-language-server

Conarius:


williamboman (approved):
Thanks! I applied some stylistic changes (YAML header and remove redundant string quotes). The `@` sign in the purl identifier needs to be [percent-encoded](https://en.wikipedia.org/wiki/Percent-encoding), otherwise everything to the right of it is parsed as the version component.

PR #4596, opened 2024-02-17
feat: add rust_hdl

Shivlocin:


williamboman (approved):
Thanks! I added darwin_arm64 target since it's not included since the latest release. I also updated the bin name to correspond to the correct name

PR #4608, opened 2024-02-18
Add zig to codelldb language list

CWood-sdf:
The Codelldb debug adapter works with zig along with c, c++, and rust. This PR makes it explicit that codelldb works with zig


williamboman (approved):
Thanks!

PR #4619, opened 2024-02-20
Add verible macOS binaries for both x86_64 and aarch64 Apple silicon

JohnnyGOX17:
Previously no support for macOS targets outside of [homebrew-verible](https://github.com/chipsalliance/homebrew-verible/tree/main). Tested binaries on M3 MacBook.


williamboman (approved):
Thanks!

PR #4648, opened 2024-02-22
feat: add fennel-ls

XeroOl:
Adds fennel-ls language server

### DIFF
--- a/packages/fennel-ls/package.yaml
+++ b/packages/fennel-ls/package.yaml
@@ -14,3 +14,7 @@ source:
 
 bin:
   fennel-ls: luarocks:fennel-ls
+
+ci_skip:
+  # https://github.com/mason-org/mason-registry/pull/4648
+  - unix
--- a/packages/fennel-ls/package.yaml
+++ b/packages/fennel-ls/package.yaml
@@ -17,4 +17,6 @@ bin:
 
 ci_skip:
   # https://github.com/mason-org/mason-registry/pull/4648
-  - unix
+  - darwin_x64
+  - linux_x64
+  - linux_x64_gnu
--- a/packages/fennel-ls/package.yaml
+++ b/packages/fennel-ls/package.yaml
@@ -1,0 +1,16 @@
+---
+name: fennel-ls
+description: Language Server for Fennel
+homepage: https://git.sr.ht/~xerool/fennel-ls
+licenses:
+  - MIT
+languages:
+  - Fennel
+categories:
+  - LSP
+
+source:
+  id: pkg:luarocks/fennel-ls@0.1.1-1
+
+bin:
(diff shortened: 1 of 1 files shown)


packages/fennel-ls/package.yaml
```diff
@@ -0,0 +1,16 @@
...
+homepage: https://git.sr.ht/~xerool/fennel-ls
+licenses:
+  - MIT
+languages:
+  - Fennel
+categories:
+  - LSP
+
+source:
+  id: pkg:luarocks/fennel-ls@0.1.1-1
```
williamboman, line 13: I see there's a new version since the PR was opened, let's update right away.

```suggestion
  id: pkg:luarocks/fennel-ls@0.1.2-2
```

Unfortunately Renovate (which is used to automate version upgrades) does not support Luarocks so this will have to manually updated in the future.

williamboman (approved):
Thanks!

XeroOl:
Oh no! I think luarocks needs to be version 3.10+ to be able to install the `fennel-ls` rock anywhere other than system-wide, due to a bug in luarocks.

williamboman:
> Oh no! I think luarocks needs to be version 3.10+ to be able to install the `fennel-ls` rock anywhere other than system-wide, due to a bug in luarocks.

Ah that's a bit unfortunate. Luarocks packages are currently only [tested on macOS](https://github.com/mason-org/actions/blob/main/tests/prepare.sh#L126-L133), which only recently updated to 3.10+. I've re-ran the test so hopefully it'll pass now.

XeroOl:
That's a strange error! I will look into it and let you know when I fix it.

williamboman:
> That's a strange error! I will look into it and let you know when I fix it.

I wasn't able to reproduce when using Luarocks installed via Homebrew. However, I was able to by replicating the installation process of the GitHub actions (basically downloading the Luarocks and Lua source tarballs, unpacking them, and compiling Luarocks against the compiled Lua artifact). Makes me wonder if there's an issue with Luarocks itself, I'll try to investigate further as well.

williamboman:
So CI is failing on `darwin_x64` even with Luarocks 3.11 installed via Homebrew. I really can't tell where the issue is, but the fact that it keeps failing when using different methods of Luarocks installation is a bit concering. Let's get this merged anyway, I'll disable CI explicitly.

XeroOl:
The issue was with the luarocks package `luarocks-build-fennel`, which fennel-ls depends on. [@tranzystorekk](https://github.com/tranzystorekk) was kind enough to make a [patch](https://lists.sr.ht/~xerool/public-inbox/patches/50304) that seems to fix the issue, and I've pushed it live to luarocks.

The CI can be enabled now.

PR #4650, opened 2024-02-22
[feat]: added systemdlint linter

magnuslarsen:
Added the `systemdlint` linter from: https://github.com/priv-kweihmann/systemdlint

The language supported here (in (n)vim terms anyway) is `systemd`, though I did not see it in the schema - do you want me to add it to the schema, or should I leave the `language` field empty?


williamboman (approved):
Thanks! I went ahead and added `systemd` as a language

PR #4702, opened 2024-02-26
Adds reformat-gherkin

conradogarciaberrotaran:


williamboman (approved):
LGTM, thanks!

packages/reformat-gherkin/package.yaml
```diff
@@ -0,0 +1,16 @@
...
+homepage: https://pypi.org/project/reformat-gherkin/
+licenses:
+  - MIT
+languages:
+  - Cucumber
+categories:
+  - Formatter
+
+source:
+  id: pkg:pypi/reformat-gherkin@3.1.0
```
williamboman, line 13: Seems like an error here - 3.0.1 seems to be the latest available version.
```suggestion
  id: pkg:pypi/reformat-gherkin@3.0.1
```

conradogarciaberrotaran:
Maybe I missed it somewhere in the documentation, but how would this tool be called from none-ls?
https://github.com/nvimtools/none-ls.nvim/discussions/100

Thanks!

PR #4740, opened 2024-02-29
feat: add cljfmt

lemonnuggets:
should close https://github.com/williamboman/mason.nvim/issues/1641

### DIFF
--- a/packages/cljfmt/package.yaml
+++ b/packages/cljfmt/package.yaml
@@ -11,7 +11,6 @@ categories:
   - Formatter
 
 source:
-  # renovate:versioning=loose
   id: pkg:github/weavejester/cljfmt@0.12.0
   asset:
     - target: darwin_arm64
@@ -20,11 +19,8 @@ source:
     - target: darwin_x64
       file: cljfmt-{{ version }}-darwin-amd64.tar.gz
       bin: cljfmt
-    - target: linux_arm64
+    - target: linux_arm64_gnu
       file: cljfmt-{{ version }}-linux-aarch64.tar.gz
-      bin: cljfmt
-    - target: linux_x64_gnu
-      file: cljfmt-{{ version }}-linux-amd64.tar.gz
       bin: cljfmt
     - target: linux_x64
       file: cljfmt-{{ version }}-linux-static-amd64.zip
--- a/packages/cljfmt/package.yaml
+++ b/packages/cljfmt/package.yaml
@@ -23,7 +23,7 @@ source:
       file: cljfmt-{{ version }}-linux-aarch64.tar.gz
       bin: cljfmt
     - target: linux_x64
-      file: cljfmt-{{ version }}-linux-static-amd64.zip
+      file: cljfmt-{{ version }}-linux-amd64-static.tar.gz
       bin: cljfmt
     - target: win_x64
       file: cljfmt-{{ version }}-win-amd64.zip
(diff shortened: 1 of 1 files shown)


lemonnuggets:
cljfmt is a clojure formatter that tries to format code while trying to make the minimal number of changes. Meanwhile zprint doesn't make that same promise while joker works on a dialect of clojure. Due to this I think cljfmt is different enough to warrant its addition into the registry.
Also cljfmt is equally popular as zprint so I don't see why cljfmt shouldn't be included if zprint is.
I'm not really familiar with how exactly I should go about making a package for mason and I just used some other packages as a reference along with what I read in the contribution guide, so if there's any concerns about the package let me know and I could see if there is anything I can do about it.

packages/cljfmt/package.yaml
```diff
@@ -0,0 +1,37 @@
...
+      bin: cljfmt
+    - target: linux_arm64
+      file: cljfmt-{{ version }}-linux-aarch64.tar.gz
+      bin: cljfmt
+    - target: linux_x64_gnu
+      file: cljfmt-{{ version }}-linux-amd64.tar.gz
+      bin: cljfmt
+    - target: linux_x64
+      file: cljfmt-{{ version }}-linux-static-amd64.zip
+      bin: cljfmt
```
williamboman, line 31: Great initiative including a GNU target! Generally static binaries are preferred, in order to avoid potential issues with libraries, specifically if the binary doesn't support "older" glibc versions. Since they provide one - let's just use that for all:

```suggestion
    - target: linux_x64
      file: cljfmt-{{ version }}-linux-static-amd64.zip
      bin: cljfmt
```

packages/cljfmt/package.yaml
```diff
@@ -0,0 +1,37 @@
...
+  # renovate:versioning=loose
+  id: pkg:github/weavejester/cljfmt@0.12.0
+  asset:
+    - target: darwin_arm64
+      file: cljfmt-{{ version }}-darwin-aarch64.tar.gz
+      bin: cljfmt
+    - target: darwin_x64
+      file: cljfmt-{{ version }}-darwin-amd64.tar.gz
+      bin: cljfmt
+    - target: linux_arm64
```
williamboman, line 23: It does however seem like the aarch64 binary is built for GNU/Linux:

```suggestion
    - target: linux_arm64_gnu
```

packages/cljfmt/package.yaml
```diff
@@ -0,0 +1,37 @@
...
+licenses:
+  - EPL-1.0
+languages:
+  - Clojure
+  - ClojureScript
+categories:
+  - Formatter
+
+source:
+  # renovate:versioning=loose
```
williamboman, line 14: No need to override the default renovate versioning (semver)
```suggestion
```

williamboman (approved):
Thanks!

packages/cljfmt/package.yaml
```diff
@@ -0,0 +1,33 @@
...
+      file: cljfmt-{{ version }}-darwin-aarch64.tar.gz
+      bin: cljfmt
+    - target: darwin_x64
+      file: cljfmt-{{ version }}-darwin-amd64.tar.gz
+      bin: cljfmt
+    - target: linux_arm64_gnu
+      file: cljfmt-{{ version }}-linux-aarch64.tar.gz
+      bin: cljfmt
+    - target: linux_x64
+      file: cljfmt-{{ version }}-linux-static-amd64.zip
```
williamboman, line 26:
```suggestion
      file: cljfmt-{{ version }}-linux-amd64-static.tar.gz
```

PR #4745, opened 2024-02-29
feat: add coffeesense

Conarius:


williamboman (approved):
Thanks! I renamed the package to `coffeesense-language-server` to better represent the underlying package/software

PR #4791, opened 2024-03-05
feat: add delance

mochaaP:


mochaaP:
cc @williamboman review request :)

mochaaP:
Rebased & updated to 2024.2.3

mochaaP:
Rebased & updated to 2024.2.106

packages/delance/package.yaml
```diff
@@ -0,0 +1,20 @@
...
+licenses:
+  - 0BSD
+  - MIT
+languages:
+  - Python
+categories:
+  - LSP
+
+source:
+  id: pkg:npm/@delance/runtime@2024.3.100
```
williamboman, line 14:
```suggestion
  id: pkg:npm/%40delance/runtime@2024.3.100
```

packages/delance/package.yaml
```diff
@@ -0,0 +1,20 @@
...
+
+schemas:
+  lsp: https://raw.githubusercontent.com/microsoft/pyright/main/packages/vscode-pyright/schemas/pyrightconfig.schema.json
```
williamboman, line 17: This seems to be a configuration schema for pyright. I couldn't find one for delance so removing
```suggestion
```

williamboman (commented):
Thanks for the PR! I was unable to find much information about this package. It seems like it's just a wrapper around the Pylance VSCode extension, and there's some [shady code that downloads artifacts](https://git.sr.ht/~self/delance-runtime/tree/trunk/item/install.mjs#L62) from the VSCode marketplace (which is against their TOS). Do you perhaps have more context?

PR #4814, opened 2024-03-06
feat: add lexical lsp support

TunkShif:
The repo(https://github.com/lexical-lsp/lexical) currently has no license file in the root directory, but is using Apache 2.0 license for some of its sub-projects located in the `projects` folder. So I'm just adding Apache 2.0 in the package license field.


williamboman (approved):
Thanks! I changed the source definition to only target `unix` because the asset only provides bash scripts (not inherently available on Windows machines)

PR #4818, opened 2024-03-06
categorise ruff as a linter and formatter

saifazmi:
ruff is also a formatter and should appear under both the Linter and Formatter category.

ref: https://docs.astral.sh/ruff/formatter/


saifazmi:
cc: @williamboman request for review please, should be a simple one :)

williamboman (approved):
Thanks!

PR #4869, opened 2024-03-09
feat() add @vue/typescript-plugin to volar

s-montigny-desautels:
Since version 2.0.0 of the @vue/language-server, @vue/typescript-plugin is required for proper TS support.

I'm not sure if it work that way, or if there is a preffered way to do this. Some guidance is welcome!

See https://github.com/vuejs/language-tools/issues/3925 for the issue in @vue/language-server repo.

Version of the @vue/language-server and @vue/typescript-plugin must match and it's a bit of a pain to maintain both version manually.

### DIFF
--- a/packages/vue-language-server/package.yaml
+++ b/packages/vue-language-server/package.yaml
@@ -13,6 +13,7 @@ source:
   id: pkg:npm/%40vue/language-server@2.0.6
   extra_packages:
     - typescript
+    - /%40vue/typescript-plugin@{{ version }}
 
 bin:
   vue-language-server: npm:vue-language-server


mehalter (approved):
This looks great. I do think that keeping the versions pinned like this makes sense for the vuejs/language-tools development process and release cycle. It looks like they do maintain the typescript-plugin and language-server in the same mono-repo and tag all of the releases at the same time with the same version.

williamboman:
It's not clear to me what the intended way to use their TypeScript integration is. My interpretation is that it's no longer bundled with the Vue language server anymore but instead is meant to be used as a plugin to `tsserver` or `typescript-language-server` separately?

It also seems like it was re-added as a dependency to the language server in https://github.com/vuejs/language-tools/pull/3944, so this would effectively do nothing?

mehalter:
You use it as a plugin with `tsserver` it's configured through the language server settings and is expected to be installed globally which makes it a good candidate for Mason. Other tools also support these language server plugins such as `typescript-tools` which would utilize this package as well

mehalter:
Oh yeah the linked PR to add the plugin as a dependency for the language server does seem like it should do enough. I can test tomorrow

s-montigny-desautels:
Also, they reverted their changes and the TS support is back in the Vue LSP Server (https://github.com/vuejs/language-tools/pull/4119). Now, it is always enabled by default and the user should opt in to be in hybrid mode. I will test this, but I think this PR can be closed.

sho-87:
> Also, they reverted their changes and the TS support is back in the Vue LSP Server ([vuejs/language-tools#4119](https://github.com/vuejs/language-tools/pull/4119)). Now, it is always enabled by default and the user should opt in to be in hybrid mode. I will test this, but I think this PR can be closed.

personally I think it'd still be useful for mason to manage typescript-plugin. its nice to have a single source manage the versions of both the plugin and volar, given they need to always match

but more than that, after testing out no hybrid mode it doesn't seem to be as usable as releases before v2 (a lot of incorrect diagnostics etc), so itd be nice to at least have the option of using hybrid mode with the plugin

fyi here is the config im currently using that feels subpar compared to hybrid mode or v1.8.7 (you have to manually disable hybrid mode with nvim): https://github.com/williamboman/mason-lspconfig.nvim/issues/371#issuecomment-2016587935

s-montigny-desautels:
An update has been made in the language-tools repo and it's clear that we don't need this PR anymore.
It was added here: https://github.com/vuejs/language-tools/pull/4134

I will close this PR, feel free to re-open it if need be.

PR #4886, opened 2024-03-11
feat: add robotcode-language-server

touchedthecode:
The package `robotframework-lsp` does not support the latest Robot Framework Version 7.0 and is currently not maintained anymore afaik. This is the alternative for Robot Framework, which is actively maintained and worked on.

Related to https://github.com/d-biehl/robotcode/issues/232


packages/robotcode-language-server/package.yaml
```diff
@@ -0,0 +1,16 @@
...
+languages:
+  - Robot Framework
+categories:
+  - LSP
+
+source:
+  id: pkg:pypi/robotcode-language-server@0.78.0
+
+bin:
+  robotframework_ls: pypi:robotcode-language-server
```
williamboman, line 16:
```suggestion
  robotcode-language-server: pypi:robotcode-language-server
```

williamboman (approved):
Thanks!

williamboman:
So it seems like the `robotcode-language-server` ships a `robotcode` executable. However, running that ends up with a 

```
ModuleNotFoundError: No module named 'robotcode.robot'
```

caused by [this import](https://github.com/d-biehl/robotcode/blob/main/src/robotcode/cli/commands/config.py#L17).

d-biehl:
hello, 

i am the developer of robotcode/robotcode language server, I have added the missing dependency to the robotcode-language-server package in version 0.78.4, so that it should run.

One thing I'm wondering is whether we need to set the version to a hardcoded version, as I update the `robotcode` fairly often. Does `package.yaml` need to be changed each time?

The second thing:

the Language Server is then used with

`robotcode language-server`, but there may still be profiles or settings that can be specified. How do you configure this?

touchedthecode:
@williamboman Sorry for missing this and thank you for catching the `ModuleNotFoundError` error!

@d-biehl Hi Daniel, thanks for fixing it! 

To answer your second question, mason just helps us install the language server, for our configuration we can/will still use https://github.com/neovim/nvim-lspconfig

williamboman:
> One thing I'm wondering is whether we need to set the version to a hardcoded version, as I update the `robotcode` fairly often. Does `package.yaml` need to be changed each time?

Renovate takes care of automatically updating versions. As long as it's not multiple releases every day it's not going to be an issue (the release cadence of your project seems perfectly fine).

> The second thing:
> 
> the Language Server is then used with
> 
> `robotcode language-server`, but there may still be profiles or settings that can be specified. How do you configure this?

Yeah I was just going to ask about this. I have only really looked at the surface of the package but to me it seems like `robotcode-language-server` has a dependency on `robotcode` which is the primary CLI interface. Installing `robotcode-language-server` only adds a subcommand to `robotcode` (there's no `robotcode-language-server` executable).

There are different ways of actually configuring the language server, the most common option is to use https://github.com/neovim/nvim-lspconfig. It does not seem like nvim-lspconfig provides a default configuration for Robotcode yet though.

**Question 1:** How do people generally interact with and install the Robotcode CLI? Installing `robotcode-language-server` through Mason will currently "override" the `robotcode` executable, which might be problematic if people rely on a certain version of `robotcode` to be installed. As such I think it'd make more sense to rename the package here to `robotcode` and primarily target the main `robotcode[all]` package, to pull in all optional dependencies:

```yaml
source:
  id: pkg:pypi/robotcode@0.78.4?extra=all
```

d-biehl:
> **Question:** How do people generally interact with and install the Robotcode CLI? Installing `robotcode-language-server` through Mason will currently "override" the `robotcode` executable, which might be problematic if people rely on a certain version of `robotcode` to be installed. As such I think it'd make more sense to rename the package here to `robotcode` and primarily target the main `robotcode[all]` package, to pull in all optional dependencies:
> 
> ```yaml
> source:
>   id: pkg:pypi/robotcode@0.78.4?extra=all
> ```

Normally you don't need to install it because the VSCode extension comes with everything you need except the robotframework package, but I think in the case of neovim we can install everything, including the debugger and so on. So we can use your approach with the `extra=all`

but 2 questions came in my mind:

1. What about the virtual Python environments? If I work in a venv environment, is the language server installed in the venv or in the system environment?

2. Is there any support in neovim for the Debug Adapter Protocol https://microsoft.github.io/debug-adapter-protocol/
    if yes, how you can use this?

williamboman:
> * What about the virtual Python environments? If I work in a venv environment, is the language server installed in the venv or in the system environment?

Packages are installed in their own isolated virtual environment. This is problematic for some packages because they rely on being able to interact with the environment, and would be better suited installed as such (i.e. not installed via Mason).

> * Is there any support in neovim for the Debug Adapter Protocol https://microsoft.github.io/debug-adapter-protocol/
>   if yes, how you can use this?

Yes there is via 3rd party plugins like https://github.com/mfussenegger/nvim-dap. It's a client implementation for DAP so it should largely function the same as for VSCode, as long as you manage to configure things properly (they have a wiki for example configurations).

williamboman:
Closing this due to installation issues (explained in https://github.com/mason-org/mason-registry/pull/4886#issuecomment-2015257166)

kimfaint:
> There are different ways of actually configuring the language server, the most common option is to use https://github.com/neovim/nvim-lspconfig. It does not seem like nvim-lspconfig provides a default configuration for Robotcode yet though.

They do now https://github.com/neovim/nvim-lspconfig/blob/12d163c5c2b05e85431f2deef5d9d59a8fd8dfc2/lua/lspconfig/configs/robotcode.lua

PR #4896, opened 2024-03-11
feat: add regols

aaronkjones:
Add support for [regols](https://github.com/kitagry/regols). Support exists in [nvim-lspconfig](https://github.com/neovim/nvim-lspconfig/blob/master/doc/server_configurations.md#regols).

### DIFF
--- a/packages/regols/package.yaml
+++ b/packages/regols/package.yaml
@@ -5,7 +5,7 @@ homepage: https://github.com/kitagry/regols
 licenses:
   - MIT
 languages:
-  - Go
+  - Rego
 categories:
   - LSP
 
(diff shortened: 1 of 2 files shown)


packages/regols/package.yaml
```diff
@@ -0,0 +1,16 @@
+---
+name: regols
+description: OPA Rego language server
+homepage: https://github.com/kitagry/regols
+licenses:
+  - MIT
+languages:
+  - Go
```
williamboman, line 8:
```suggestion
  - Rego
```

williamboman (approved):
Thanks!

PR #4938, opened 2024-03-13
feat(php_codesniffer): update repository location

Rasmus-Bertell:
PHP_CodeSniffer development is abandoned in squizlabs repository (https://github.com/squizlabs/PHP_CodeSniffer), but continues under PHPCSStandards repository (https://github.com/PHPCSStandards/PHP_CodeSniffer).

This PR only updates the source and doesn't update the version, squizlabs is still on version 3.7.2 but the updated repository is on version 3.9.0. I chose to not update the version to minimize the possible impact of this PR.


williamboman (approved):
Thanks!

PR #4944, opened 2024-03-14
bump(haskell-language-server): update to version 2.7.0.0

shenawy29:
I updated hls to the latest and recommended version


williamboman (commented):
Thanks!

PR #4945, opened 2024-03-14
feat: add erb-formatter

saifazmi:
Hi, just wanted to add a formatter for Ruby's ERB template files.

Thanks for considering this addition.

---

gem: https://rubygems.org/gems/erb-formatter
ERB ref: https://github.com/ruby/erb

### DIFF
--- a/packages/erb-formatter/package.yaml
+++ b/packages/erb-formatter/package.yaml
@@ -1,0 +1,17 @@
+---
+name: erb-formatter
+description: Format ERB files with speed and precision.
+homepage: https://github.com/nebulab/erb-formatter
+licenses:
+  - MIT
+languages:
+  - HTML
+  - Ruby
+categories:
+  - Formatter
+
+source:
+  id: pkg:gem/erb-formatter@0.7.2
+
+bin:
+  erb-format: gem:erb-formatter
--- a/packages/erb-formatter/package.yaml
+++ b/packages/erb-formatter/package.yaml
@@ -14,4 +14,4 @@ source:
   id: pkg:gem/erb-formatter@0.7.2
 
 bin:
-  erb-format: gem:erb-formatter
+  erb-formatter: gem:erb-formatter


saifazmi:
cc @williamboman all checks are green, should be an easy one to merge.

packages/erb-formatter/package.yaml
```diff
@@ -0,0 +1,17 @@
...
+  - HTML
+  - Ruby
+categories:
+  - Formatter
+
+source:
+  id: pkg:gem/erb-formatter@0.7.2
+
+bin:
+  erb-format: gem:erb-formatter
```
williamboman, line 17:
```suggestion
  erb-formatter: gem:erb-formatter
```

williamboman (approved):
Thanks!

packages/erb-formatter/package.yaml
saifazmi: Hi @williamboman can you please revert this to `erb-format` or is there a way that we can have 2 binaries `erb-format` and `erb-formatter`?

I understand that there might be some confusion because of the difference between the name of the gem and its corresponding binary, but `erb-format` is the correct way to use this gem as documented [here](https://github.com/nebulab/erb-formatter?tab=readme-ov-file#usage).

And other nvim tools like `conform.nvim` are configured to use the `erb-format` binary, as you can see [here](https://github.com/stevearc/conform.nvim/blob/master/lua/conform/formatters/erb_format.lua).

Thanks :)

williamboman: Ah, apologies. It's a common thing that the executable name is incorrect so it's a habit, should've double checked. That seems really odd though. I guess they provide both `erb-format` and `erb-formatter` executables? 7'll update right away

williamboman: Will be fixed in https://github.com/mason-org/mason-registry/pull/5044. They do provide both executables, I guess the latter is an old variant.

saifazmi: They do provide both variants, and I guess you are right it must be an old executable and old documentation, and everyone else just ran with it in their tools. Good old legacy issue :)

Thanks for sorting this out, much appreciated.

PR #4957, opened 2024-03-14
Add spring boot language server schema package

asmodeus812:
Initial attempt at adding the resources for the spring boot language server. The language server itself is presented as an exploded jar, so it does really have a jar binary, in other words it has to be started with `java -cp <classpath> <mainclass>`. More information can be found here.

https://github.com/spring-projects/sts4/issues/1128#issuecomment-1872665406
https://github.com/spring-projects/sts4/blob/8bf29f4b9126135e4d5eb4ab7c02136f4ef3c9ba/vscode-extensions/commons-vscode/src/launch-util.ts#L231w
https://github.com/emacs-lsp/lsp-java/blob/master/lsp-java-boot.el#L128

### DIFF
--- a/packages/springboot-language-server/package.yaml
+++ b/packages/springboot-language-server/package.yaml
@@ -1,0 +1,21 @@
+---
+name: springboot-language-server
+description: SpringBoot Language Server.
+homepage: https://github.com/spring-projects/sts4
+licenses:
+  - EPL-1.0
+languages:
+  - XML
+  - YAML
+  - Java
+categories:
+  - Linter
+
+source:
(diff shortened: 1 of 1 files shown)


williamboman (approved):
Thanks for the PR! Do you have an example of how to actually get this running inside Neovim, using the contents of that release asset?

asmodeus812:
Still working on some basic mvp. Using the vscode impl as s reference. But when i do will try to contribute those into mason lspconfig, there are some mason specific resources required,  and the more generic / basic config in  nvim lspconfig

TheRustifyer:
@asmodeus812 friendly ping. Is this alive?

PR #4961, opened 2024-03-14
Add entry for circleci-yaml-language-server

ELD:


williamboman (approved):
Thanks! It seems like the linux_* binaries are built for GNU so I made some adjustments. I also added the `schema.json` to `$MASON/share/circleci-yaml-language-server/schema.json` to be used when starting the client, for example:

```lua
vim.lsp.start {
  name = "cci",
  cmd = { "circleci-language-server", "--stdio" },
  cmd_env = {
    SCHEMA_LOCATION = vim.fn.expand "$MASON/share/circleci-yaml-language-server/schema.json"
  }
}
```

The VSCode extension [seems to do a fair bit of further initialization on the client side](https://github.com/CircleCI-Public/circleci-yaml-language-server/blob/main/editors/vscode/src/server.ts#L215-L257), not sure how necessary they are for the LSP to function reasonably well.

PR #4994, opened 2024-03-17
Add ruby config schema

iseth:
add ruby-lsp config schema

### DIFF
--- a/packages/ruby-lsp/package.yaml
+++ b/packages/ruby-lsp/package.yaml
@@ -13,5 +13,8 @@ categories:
 source:
   id: pkg:gem/ruby-lsp@0.14.6
 
+schemas:
+  lsp: vscode:https://raw.githubusercontent.com/Shopify/ruby-lsp/main/vscode/package.json
+
 bin:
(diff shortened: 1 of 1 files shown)


williamboman (commented):
Thanks for the PR! It seems like that configuration schema only contains settings related to the VSCode client, but I might be missing something. Are there any particular reason you've added it?

iseth:
Correct me if i am wrong but that schema has lsp command that i wanted to be able to change / toggle in mason. 
Like code formatting or code actions etc...

williamboman:
> Correct me if i am wrong but that schema has lsp command that i wanted to be able to change / toggle in mason. Like code formatting or code actions etc...

Ah this is currently only used for visualization in the `:Mason` window. Mason itself isn't involved in starting or configuring servers in any way, it only provides the packages.

If you use nvim-lspconfig you'll probably want to do something like:

```lua
require("lspconfig").ruby_ls.setup {
  init_options = {
    enabledFeatures = {
      codeAction = false
    }
  }
}
```

PR #4997, opened 2024-03-17
feat: add cadence

Conarius:


Conarius:
@williamboman
had to remove the bin part as it uses the flow-cli package as a dependency which can't be installed through npm.

williamboman (commented):
Thanks for the PR! How is this meant to be used?

Conarius:
@williamboman
When you have installed the flow-cli and then the cadence lsp you can use the cadence lsp via `flow cadence language-server`.

Conarius:
> > When you have installed the flow-cli and then the cadence lsp you can use the cadence lsp via `flow cadence language-server`.
> 
> I don't believe this package is meant to be installed directly by users? It seems like the Cadence language server is bundled with the flow cli itself? I am able to start it successfully via `flow cadence language-server`.

Yeah, it is a bit odd how they do it. Seems to be bundled into the flow cli itself

williamboman:
Closing this as it seems to be incorrect installation/usage of the package, thanks for the PR anyways!

PR #5000, opened 2024-03-17
feat: add buck2

Conarius:


williamboman:
Hello and thanks for the PR (and apologies for late review)! This seems<sup>1</sup> to be a package that exceeds the scope of the core registry. An example of such a package is something that installs an entire language toolchain or platform through either a single or multiple binaries. Management of such packages through Mason has a history of causing issues/confusion for both users and the maintainers of the software themselves. You can always host your own packages through your own registry, see https://github.com/williamboman/mason-registry-playground/ for a starting point.

<sup>1 - please let me know if you disagree with my assessment</sup>

PR #5013, opened 2024-03-18
feat: add cobol ls

Conarius:


packages/cobol-ls/package.yaml
```diff
@@ -0,0 +1,29 @@
...
+      file: cobol-language-support-linux-arm64-{{version}}.vsix
+    - target: linux_x64
+      file: cobol-language-support-linux-x64-{{version}}.vsix
+    - target: win_arm64
+      file: cobol-language-support-win32-arm64-{{version}}-signed.vsix
+    - target: win_x64
+      file: cobol-language-support-win32-x64-{{version}}-signed.vsix
+
+bin:
+  cobol-language-support: "{{source.asset.bin}}"
```
williamboman, line 29: `{{source.asset.bin}}` gets expanded to `nil` because it's not defined above, resulting in no bin being linked. Do they have a recommended way of running the server? It seems like the assets contain compiled Java bytecode as well as a platform-specific GraalVM binary

Conarius: It also has a native binary found in extension/server/native. Maybe I should add that path to the source

Conarius:
@williamboman
Just added the missing binary paths so it should be ready. Though only the windows version needs a dll that is shipped in the vsix package

williamboman (approved):
Thanks!

PR #5021, opened 2024-03-19
chore: update ruff's description to include formatter

saifazmi:
Missed out on updating the description in the previous PR https://github.com/mason-org/mason-registry/pull/4818


williamboman (approved):
Thansk!

PR #5028, opened 2024-03-19
feat: add dcm

Conarius:


Conarius:
> Thanks! This seems to be a formatter and linter as well, no? In that case we should expand the categories list

Yeah, this is a (commercial) linter and formatter as well for some reason

PR #5056, opened 2024-03-21
Adds ginko registry entry

SethGower:
## Describe your changes
Adds a registry entry for [`ginko_ls`](https://github.com/Schottkyc137/ginko), which is a Device Tree (DTS) Parser and language server. It's very early stages right now, but it's functional. I will also be working on a PR in `nvim-lspconfig` to have a built in configuration for this, as I am currently using a custom defined one.

I tested with the example that the Ginko readme provides. 
```dts
/dts-v1/;

/ {
    pic@10000000 {
        phandle = <1>;
        interrupt-controller;
        reg = <0x10000000 0x100>;
    }
};
```

In order to configure the LSP, I added this to my nvim configs
```lua
    if not configs.ginko_ls then
        configs.ginko_ls = {
            default_config = {
                cmd = { "ginko_ls" };
                filetypes = { "dts" };
                root_dir = function (_)
                    return vim.fs.dirname(vim.fs.find({ '.git' }, { upward = true })[1]);
                end,
                settings = {};
            };
        }
    end
```
Which showed an error after the `}` for `pic@10000000` node.
## Screenshots
image
Screenshot showing that it is appropriately using the `ginko_ls` binary in the mason path.

### DIFF
--- a/packages/ginko_ls/package.yaml
+++ b/packages/ginko_ls/package.yaml
@@ -17,7 +17,7 @@ source:
     - target: darwin_arm64
       file: ginko_ls-aarch64-apple-darwin.zip
       bin: ginko_ls-aarch64-apple-darwin/bin/ginko_ls
-    - target: linux_x64
+    - target: linux_x64_gnu
       file: ginko_ls-x86_64-unknown-linux-gnu.zip
       bin: ginko_ls-x86_64-unknown-linux-gnu/bin/ginko_ls
     - target: win_x64
--- a/packages/ginko_ls/package.yaml
+++ b/packages/ginko_ls/package.yaml
@@ -3,7 +3,7 @@ name: ginko_ls
 description: |
   ginko_ls is meant to be a feature-complete language server for device-trees. Language servers can be used in many
   editors, such as Visual Studio Code, Emacs or Vim
-homepage: https://github.com/Schottkyc137
+homepage: https://github.com/Schottkyc137/ginko
 licenses:
   - MIT
 languages:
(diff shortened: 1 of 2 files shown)


packages/ginko_ls/package.yaml
```diff
@@ -0,0 +1,28 @@
...
+categories:
+  - LSP
+
+source:
+  id: pkg:github/Schottkyc137/ginko@v0.0.4
+  asset:
+    - target: darwin_arm64
+      file: ginko_ls-aarch64-apple-darwin.zip
+      bin: ginko_ls-aarch64-apple-darwin/bin/ginko_ls
+    - target: linux_x64
```
williamboman, line 20:
```suggestion
    - target: linux_x64_gnu
```

williamboman (approved):
Thanks!

williamboman:
> FWIW this could also be built from `cargo` so not sure which route is preferable. I figured the released binaries route was better since it's faster, and more reliable with respect to if there is an issue, you can just tell the author of ginko which version is installed from the releases.

Yeah prebuilt release assets are preferred. I saw that the x64 Linux binary is built for GNU/Linux so I updated the target

SethGower:
Awesome, thanks! Any other things you would like addressed?  I realize it's a simple PR, but still. 

Also, just wanted to say great job with this plugin, it's fantastic and makes things _so easy_.

packages/ginko_ls/package.yaml
```diff
@@ -0,0 +1,28 @@
+---
+name: ginko_ls
+description: |
+  ginko_ls is meant to be a feature-complete language server for device-trees. Language servers can be used in many
+  editors, such as Visual Studio Code, Emacs or Vim
+homepage: https://github.com/Schottkyc137
```
williamboman, line 6:
```suggestion
homepage: https://github.com/Schottkyc137/ginko
```

PR #5175, opened 2024-03-30
Support markdown oxide binaries

Feel-ix-343:
Are the platforms typical? How does this look at @williamboman


Feel-ix-343:
Hey I am also noticing that the binaries are a significantly slower than compiling from source. Is this typical?

Feel-ix-343:
I fixed it by changing the linux rustc target. Very fast now

Feel-ix-343:
Hey @williamboman I am a highschooler trying to update my waitlist colleges with this project. Rigth now, nobody without rust can use it, and this is blocking to my publishing of the project. I really want to share a great update with stats and details from publishing the project, but the deadline is in about a week. Would you mind reviewing this soon so I can publish?

Thank you

williamboman (approved):
Thanks!

PR #5387, opened 2024-04-17
feat(pylyzer): use github releases instead of `cargo`

chrisgrieser:
## Describe your changes
Use github releases instead of `cargo`. This has the advantage that users do not require the rust toolchain to install this package.


williamboman (approved):
Thanks!

PR #5395, opened 2024-04-18
add earthlyls configuration

glehmann:
## Describe your changes
add the configuration to download earthlyls binaries from the github release

related to https://github.com/neovim/nvim-lspconfig/pull/3111

### DIFF
--- a/packages/earthlyls/package.yaml
+++ b/packages/earthlyls/package.yaml
@@ -26,7 +26,7 @@ source:
       file: earthlyls-{{ version }}-macos-arm64.tar.gz
       bin: earthlyls-{{ version }}-macos-arm64/earthlyls
     - target: win_x64
-      file: earthlyls-{{ version }}-windows-amd64.tar.gz
+      file: earthlyls-{{ version }}-windows-amd64.zip
       bin: earthlyls-{{ version }}-windows-amd64/earthlyls.exe
 
 bin:
--- a/packages/earthlyls/package.yaml
+++ b/packages/earthlyls/package.yaml
@@ -1,0 +1,34 @@
+---
+name: earthlyls
+description: |
+  A fast language server for earthly.
+homepage: https://github.com/glehmann/earthly
+licenses:
+  - MIT
+languages:
+  - earthfile
+categories:
+  - LSP
+
+source:
+  # renovate:versioning=loose
+  id: pkg:github/glehmann/earthlyls@0.1.2
+  asset:
+    - target: linux_x64
+      file: earthlyls-{{ version }}-linux-amd64.tar.gz
+      bin: earthlyls-{{ version }}-linux-amd64/earthlyls
+    - target: linux_arm64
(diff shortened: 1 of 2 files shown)


glehmann:
I may be able to provide the binaries for other platform, but I can’t find what `target` to use. Could you point me to a list of supported target or to how I can build them?

packages/earthlyls/package.yaml
```diff
@@ -0,0 +1,34 @@
+---
+name: earthlyls
+description: |
+  A fast language server for earthly.
+homepage: https://github.com/glehmann/earthly
```
williamboman, line 5: I believe the intended URL is:
```suggestion
homepage: https://github.com/glehmann/earthlyls
```

schemas/enums/language.json
```diff
@@ -50,6 +50,7 @@
         "Docker",
         "Drools",
         "DTS",
+        "Earthly",
```
williamboman, line 53: I'd suggest using "Earthly" as the language entry. While technically not a language, these entries are primarily meant to allow users to discover relevant tooling and I believe Earthly would be the most appropriate one in this case.

packages/earthlyls/package.yaml
```diff
@@ -0,0 +1,34 @@
...
+licenses:
+  - MIT
+languages:
+  - Earthly
+categories:
+  - LSP
+
+source:
+  # renovate:versioning=loose
```
williamboman, line 14: The project seems to apply semver versioning, in which case we don't need to override the default versioning strategy.
```suggestion
```

williamboman:
> I may be able to provide the binaries for other platforms, but I can’t find what `target` to use. Could you point me to a list of supported target or to how I can build them?

The full list of supported platforms can be found [here](https://github.com/mason-org/mason-registry/blob/main/schemas/enums/platform.json#L5-L35). The primary ones are `darwin_x64`, `darwin_arm64`, `linux_x64`, `linux_arm64` and `win_x64`, so this LGTM already!

packages/earthlyls/package.yaml
```diff
@@ -0,0 +1,33 @@
...
+      file: earthlyls-{{ version }}-linux-arm64.tar.gz
+      bin: earthlyls-{{ version }}-linux-arm64/earthlyls
+    - target: darwin_x64
+      file: earthlyls-{{ version }}-macos-amd64.tar.gz
+      bin: earthlyls-{{ version }}-macos-amd64/earthlyls
+    - target: darwin_arm64
+      file: earthlyls-{{ version }}-macos-arm64.tar.gz
+      bin: earthlyls-{{ version }}-macos-arm64/earthlyls
+    - target: win_x64
+      file: earthlyls-{{ version }}-windows-amd64.tar.gz
```
williamboman, line 29:
```suggestion
      file: earthlyls-{{ version }}-windows-amd64.zip
```

williamboman (approved):
LGTM! If you'd like to expand the asset list with more platforms we can do so in later PR

PR #5454, opened 2024-04-23
feat: add crlfmt configuration

rail:
## Describe your changes
Adding the `crlfmt` formatter, a `gofmt`-style formatter for Go code that enforces the [CockroachDB Style Guide](https://wiki.crdb.io/wiki/spaces/CRDB/pages/181371303/Go+coding+guidelines).

## Screenshots

image

### DIFF
--- a/packages/crlfmt/package.yaml
+++ b/packages/crlfmt/package.yaml
@@ -1,0 +1,17 @@
+---
+name: crlfmt
+description: Formatter for Go code that enforces the CockroachDB Style Guide.
+homepage: https://github.com/cockroachdb/crlfmt
+licenses:
+  - Apache-2.0
+  - BSD-3-Clause
+languages:
+  - Go
+categories:
+  - Formatter
+
+source:
+  id: pkg:golang/github.com/cockroachdb/crlfmt@v0.0.0-20230505164321-461e8663b4b4
+
+bin:
+  crlfmt: golang:crlfmt


williamboman (approved):
Thanks! The CI will fail because `v0.0.0-20230505164321-461e8663b4b4` is not a recognized version according to the Go Pkg API, which causes the test suite to fail. There are other Go packages with this issue with pending PRs, I'll need to figure out the best way to bypass version validation for untagged Go packages.

rail:
Oh, interesting.
I generated the version by running the following command:
```
go list -f '{{.Version}}' -m github.com/cockroachdb/crlfmt@master
```

To verify it works, I created a temp registry release, containing only the package above. You can see the release files here: https://github.com/rail/mason-registry/releases/tag/crlfmt

Then to verify it installs correctly I added the following entry in the lazy setup:
```
  {
    "williamboman/mason.nvim",
    opts = {
      registries = {
        "github:rail/mason-registry@crlfmt",
        "github:mason-org/mason-registry",
      },
    },
  },
```

So far it works as expected. I wonder how CI is different. 🤔

williamboman:
Yeah installation of the package will work fine, it's just how mason.nvim gathers available versions for it (CI runs in "strict" mode which will cause installation to fail early if the provided version is not available upstream). Upstream in this case is the https://proxy.golang.org/ API: https://proxy.golang.org/github.com%2Fcockroachdb%2Fcrlfmt/@v/list, which returns nothing (compared to e.g. https://proxy.golang.org/golang.org%2Fx%2Ftools%2Fgopls/@v/list).

williamboman:
Yes! Tagged versions are very much preferred than tracking HEAD

rail:
Here we go
```
$ curl https://proxy.golang.org/github.com%2Fcockroachdb%2Fcrlfmt/@v/list
v0.1.0
```

williamboman (approved):
Great, thanks!

rail:
Thank you very much!

PR #5500, opened 2024-04-26
bump-fix(zls): update to v0.12.0

qazo:
## Describe your changes
Updated version to 0.12.0

- looks like the zls binaries are no longer in a bin subdir but at the root of the archive.
- Also added an overrride for 0.11.0
- archive format for linux and macos has changed to `tar.xz`


williamboman (approved):
Thanks for also adding a version override!

PR #5689, opened 2024-05-11
feat: add shopify-cli

slavamak:
## Describe your changes
Adds [`shopify-cli`](https://github.com/Shopify/cli) tool to help create and work with Shopify themes. This will be very useful for better integration via nvim-lspconfig. Currently we have the [`theme_check`](https://github.com/mason-org/mason-registry/blob/main/packages/shopify-theme-check/package.yaml) package which is considered deprecated. Although I haven't found any information about it. But you can see that there is no activity in the [repository](https://github.com/Shopify/theme-check) for the last year, also `shopify-cli` basically includes theme_check (linter) and lsp. Thank you!


williamboman (approved):
Thanks!

PR #5699, opened 2024-05-12
Added support for Swift LSP

Santhosh-KS:
## Describe your changes
Added a Swift LSP support in neovim. Started working with Swift code in Linux using neovim. Found that there was no support for LSP binidings in Mason, hence took some time to introduce sourcekit-lsp for Swift programming language. There exists a VS code extension maintained by SSWC, I'm creating required mason-registry to pull it from the VSX repository.

- [ x] I have successfully tested installation of the package.
- [ x] I have successfully tested the package after installation.

### DIFF
--- a/packages/sourcekit/package.yaml
+++ b/packages/sourcekit/package.yaml
@@ -1,0 +1,28 @@
+---
+name: sourcekit-lsp
+description: |
+  This extension adds language support for Swift to neovim IDE. It supports:
+    Code completion
+    Jump to definition, peek definition, find all references, symbol search
+    Error annotations and apply suggestions from errors
+    Automatic generation of launch configurations for debugging with CodeLLDB
+    Automatic task creation
+    Package dependency view
+    Test Explorer view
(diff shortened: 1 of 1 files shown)


Santhosh-KS:
Hi,
I found one issue while testing. I don't see any discussion forum to get answers. Please let me know how to go about it?
Question:
When I used the local package it is installing. But when I open a swift file and do "LspInfo" there are no clients attahed. After further investingation I figured that there is an error I'm getting when i try to run "sourckit" locally manuall regarding "Error: nomodule named vscode" any inputs on how to solve these errors? This is coming from the vsix package I'm downloading. Looks like not all the dependencies are not packaged as part of the vsix file. How to put some preconditions for this installations?

Conarius:
I think the schema should also be added if one is available

Santhosh-KS:
Done. Added Schema.

williamboman (commented):
Hello and thanks for the PR! This seems to be a pure VSCode extension that only works inside VSCode. Furthermore, sourcekit-lsp relies on outputs from tools in your current toolchain so it's not probably not a good candidate for addition to Mason.

PR #5720, opened 2024-05-14
feat: add lazydocker

varbhat:
[lazydocker](https://github.com/jesseduffield/lazydocker) is the lazier way to manage everything docker.


williamboman (commented):
Hello and thanks for the PR! I'm afraid this package is out of scope for the core registry at the moment

PR #5721, opened 2024-05-14
feat: add k9s

varbhat:
[k9s](https://github.com/derailed/k9s) is Kubernetes CLI To Manage Your Clusters In Style!


williamboman (commented):
Hello and thanks for the PR! I'm afraid this package is out of scope for the core registry at the moment

PR #5865, opened 2024-05-25
fix(sqls): add `schemas.lsp` entry

nekowinston:
## Describe your changes
The `sqls` package was missing an LSP schema url that was readily available in the source repository.


williamboman (approved):
Thanks!

PR #5866, opened 2024-05-25
fix(some-sass): pin `schemas.lsp` url

nekowinston:
## Describe your changes
I double-checked the LSP schema URLs, and it'd seem to me that the [repo structure on some-sass](https://github.com/wkillerud/some-sass/blob/some-sass-language-server@1.2.3/vscode-extension/package.json) would allow version pinning.


williamboman (approved):
Thanks!

PR #5867, opened 2024-05-25
fix(vtsls): pin `schemas.lsp` url

nekowinston:
## Describe your changes
Same as #5866 but for vtsls


williamboman (approved):
Thanks!

PR #5871, opened 2024-05-25
feat: add syntax-tree

Conarius:
## Describe your changes
adds syntax tree as supported by nvim-lspconfig


williamboman (approved):
Thanks! Perhaps a nitpick but I renamed to `stree` because `syntax-tree` doesn't seem to be used to refer to this package, and it's a bit generic in terms of naming.

PR #5904, opened 2024-05-29
fix(gleam): update deprecation message with more context

justinrassier:
## Describe your changes
The Gleam's LSP is bundled with the rest of its toolchain. To avoid version conflicts, it is recommended to not have Mason manage the LSP. This resulted in an original deprecation message that was added. The message could use a bit more context as to not confuse those getting started with Gleam.

## Screenshots
image

### DIFF
--- a/packages/gleam/package.yaml
+++ b/packages/gleam/package.yaml
@@ -12,7 +12,7 @@ categories:
   - LSP
 
 deprecation:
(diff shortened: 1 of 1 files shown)


justinrassier:
closed in favor of https://github.com/mason-org/mason-registry/pull/5911 to remove Gleam from the registry all together.

PR #5911, opened 2024-05-29
fix(gleam): remove gleam from the registry

justinrassier:
## Describe your changes

The Gleam's LSP is bundled with the rest of its toolchain. To avoid version conflicts, it is recommended to not have Mason manage the LSP. This resulted in an original deprecation message that was added. Even with this message, it is causing confusion to people coming to the language.

The creator of the language has asked to see if we could remove Gleam from the registry all together. I have tested this out locally and confirmed that the Gleam language doesn't show up inside of the listing.

Replaces https://github.com/mason-org/mason-registry/pull/5904


lpil:
Hi, I'm the lead developer of Gleam.

It would be great if we could remove this package as even though it is deprecated we are getting very regular support requests and people posting online that the Gleam language server does not work, and upon further inspection is that they have not seen or have ignored the deprecation message and have installed it anyway.

I don't know how we can stop this problem from happening beyond removing the package. I would be very grateful is there's anything you can do to help us here, as the package is consistently causing us problems and damaging our reputation.

lpil:
Thank you 💜

PR #5924, opened 2024-06-01
feat(astro): add `@astrojs/ts-plugin` as extra package to `astro-language-server`

folke:
## Describe your changes
Support for importing astro files in a typescript file needs to be setup by adding the `@astrojs/ts-plugin` as a plugin to `tsserver` or `vtsls`.

The `@astrojs/ts-plugin` has been added as an extra package, so that it's easy to add this to `vtsls`.

See https://github.com/withastro/language-tools/tree/main/packages/ts-plugin

The plugin is part of the overall astro language tools that also includes the astro language server.

A similar thing is needed for vue projects, but luckily the ts plugin is already part of the volar package.

### DIFF
--- a/packages/astro-language-server/package.yaml
+++ b/packages/astro-language-server/package.yaml
@@ -13,6 +13,7 @@ source:
   id: pkg:npm/%40astrojs/language-server@2.10.0
   extra_packages:
     - typescript
+    - "@astrojs/ts-plugin"
 
 schemas:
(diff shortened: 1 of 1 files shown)


folke:
Is there anything blocking this? No rush, but if this is not something you're interested in, or if you'd like to do this differently, then it would be great to know :)

LazyVim's astro and svelte support need this (or something like this).

Link to the other PR https://github.com/mason-org/mason-registry/pull/5925

williamboman (approved):
Thanks!

folke:
Thanks for merging!

Issue #5922, opened 2024-05-30
LemMinX [0.27.0] - OSX (Intel): Bad CPU


alycklama:
I'm getting a `bad CPU type in executable: ./lemminx` when executing 0.27.0 installed from Mason.

0.26.1 works fine, downloaded from the release page:
https://github.com/redhat-developer/vscode-xml/releases/

alycklama:
I noticed this is already an issue with the official binary.

Opened a new issue here:
https://github.com/redhat-developer/vscode-xml/issues/1001

PR #5944, opened 2024-06-01
fix(lemminx): split macOS targets

radimsuckr:
## Describe your changes
In regards to https://github.com/mason-org/mason-registry/issues/5922 and https://github.com/redhat-developer/vscode-xml/issues/1001, the same issue also occurs on macOS with ARM64 chips. The error I get on ARM64 is:

The current machine does not support all of the following CPU features that are required by the image: [CX8, CMOV, FXSR, MMX, SSE, SSE2, SSE3, SSSE3, SSE4_1, SSE4_2, POPCNT, LZCNT, AVX, AVX2, BMI1, BMI2, FMA]. Please rebuild the executable with an appropriate setting of the -march option.

Splitting the targets and installing the correct artifact fixes this. I tested it locally. :)


williamboman (approved):
Thanks!

PR #5952, opened 2024-06-02
feat: add fish-lsp

Conarius:
## Describe your changes
adding fish-lsp as it is recently supported by nvim-lspconfig.
Though fish and node version 21.7.1 are dependencies (!)


williamboman:
Hey and thanks for the PR (and sorry for late review)! Custom build scripts are being deprecated because of difficulty in maintaining them over time and their common reliance on system dependencies that Mason cannot detect. You can always host your own registry if you'd like to manage it through Mason, see https://github.com/williamboman/mason-registry-playground/ for a starting point

PR #5968, opened 2024-06-03
bump(sonarlint-language-server): update to v4.6.0

thahnen:

I've noticed I forgot to update it for the three last releases, too much to do :disappointed:


williamboman (approved):
Thanks!

PR #5862, opened 2024-05-25
feat: add tsp-server

Conarius:
## Describe your changes
Adds tsp Server for typespec files recently added into nvim-treesitter

## Checklist before requesting a review

- [x] I have successfully tested installation of the package.
- [x] I have successfully tested the package after installation.

PR #6021, opened 2024-06-06
added support for typespec lsp

jpx40:
## Describe your changes

Hey, 
i added the typespec lsp pkg. 
the lsp is bundled with the typespec compiler so it will install both.
The LSP bin is tsp-server.


Conarius:
Two minds think alike xD
I already made a pr for tsp-server

jpx40:
> Two minds think alike xD I already made a pr for tsp-server

think you was faster

williamboman:
Thanks for the PR! Added in #5862

Issue #6046, opened 2024-06-08
Add: agda


21st-centuryman:
I want to add the agda programming language to mason

https://github.com/agda/agda-language-server

foxyseta:
Duplicate of #3936?

PR #5263, opened 2024-04-05
Feat: Agda Language Server

foxyseta:
## Describe your changes

Add Agda-Language-Server. It appears the VSCode package manifest uses yaml rather than json, which seems not to be supported by Mason apparently.

## Issue ticket number and link
Closes #3936.
Old PR: #3972

## Checklist before requesting a review

- [x] I have successfully tested installation of the package.
- [x] I have successfully tested the package after installation.

PR #6047, opened 2024-06-08
Add: agda

21st-centuryman:
## Describe your changes
Added the agda language to mason

#6046

### DIFF
--- a/packages/agda-language-server/package.yaml
+++ b/packages/agda-language-server/package.yaml
@@ -1,0 +1,26 @@
+---
+name: agda-language-server
+description: Language Server for Agda.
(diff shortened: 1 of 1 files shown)


foxyseta:
I previously opened https://github.com/mason-org/mason-registry/pull/5263 which is also for the Agda Language Server, but where `stack install` is used instead. This is because the ALS's Linux releases are exclusively for Ubuntu-based distros. (I am personally fine with merging either @williamboman).

21st-centuryman:
Ah ok I must have missed it. I am also fine with closing this if `stack install` is prefered.

PR #6057, opened 2024-06-09
feat(neocmakelsp): use github release instead of building from source

uyha:

### DIFF
--- a/packages/neocmakelsp/package.yaml
+++ b/packages/neocmakelsp/package.yaml
@@ -10,7 +10,18 @@ categories:
   - LSP
 
 source:
-  id: pkg:cargo/neocmakelsp@0.7.9
+  id: pkg:github/Decodetalkers/neocmakelsp@v0.7.9
+  asset:
+    - target: darwin_arm64
+      file: neocmakelsp-aarch64-apple-darwin
+    - target: darwin_x64
+      file: neocmakelsp-x86_64-apple-darwin
+    - target: win_x64
+      file: neocmakelsp-x86_64-pc-windows-msvc.exe
+    - target: linux_x64_gnu
+      file: neocmakelsp-x86_64-unknown-linux-gnu
+    - target: linux_x64_musl
+      file: neocmakelsp-x86_64-unknown-linux-musl
 
 bin:
-  neocmakelsp: cargo:neocmakelsp
+  neocmakelsp: "{{source.asset.file}}"


Conarius:
What would be the benefit of it other than faster installation?
Cuz the releases are only for certain platforms and sometimes the release assets change so there's that

uyha-kwz:
it doesn’t require users to have the rustc and cargo installed in order to use this.

uyha:
it doesn’t require users to have `rustc` and `cargo` installed in order to use this.

packages/neocmakelsp/package.yaml
```diff
@@ -10,7 +10,16 @@ categories:
...
 source:
-  id: pkg:cargo/neocmakelsp@0.7.3
+  id: pkg:github/Decodetalkers/neocmakelsp@v0.7.3
+  asset:
+    - target: darwin_arm64
+      file: neocmakelsp-aarch64-apple-darwin
+    - target: darwin_x64
+      file: neocmakelsp-x86_64-apple-darwin
+    - target: linux_x64_gnu
+      file: neocmakelsp-x86_64-unknown-linux-gnu
```
williamboman, line 20: We probably want to target the `-musl` variant for broader support (can remove the `_gnu` specifier then).

uyha: Upstream has both gnu and musl, I updated the PR to support both

packages/neocmakelsp/package.yaml
```diff
@@ -10,7 +10,18 @@ categories:
   - LSP
 
 source:
-  id: pkg:cargo/neocmakelsp@0.7.9
+  id: pkg:github/Decodetalkers/neocmakelsp@v0.7.9
```
eeeXun, line 13: The repo name changes to `github/neocmakelsp/neocmakelsp`

packages/neocmakelsp/package.yaml
```diff
@@ -10,7 +10,18 @@ categories:
...
+    - target: darwin_arm64
+      file: neocmakelsp-aarch64-apple-darwin
+    - target: darwin_x64
+      file: neocmakelsp-x86_64-apple-darwin
+    - target: win_x64
+      file: neocmakelsp-x86_64-pc-windows-msvc.exe
+    - target: linux_x64_gnu
+      file: neocmakelsp-x86_64-unknown-linux-gnu
+    - target: linux_x64_musl
+      file: neocmakelsp-x86_64-unknown-linux-musl
```
eeeXun, line 25: It supports aarch64 for linux since [v0.8.12.2](https://github.com/neocmakelsp/neocmakelsp/releases/tag/v0.8.12.2).

packages/neocmakelsp/package.yaml
uyha: The mason maintainer doesn't seem to care about this PR so I stop pursuing, I can close it and let you take over if you want.

eeeXun: > The mason maintainer doesn't seem to care about this PR so I stop pursuing, I can close it and let you take over if you want.

Sure. I would love to.

PR #6217, opened 2024-06-22
fix(zls): use lsp config schema from ZLS instead of vscode-zig

Techatrix:
## Describe your changes

The LSP config options for [ZLS](https://github.com/zigtools/zls) are currently provided by using the [package.json](https://github.com/ziglang/vscode-zig/blob/master/package.json) from the [vscode-zig](https://github.com/ziglang/vscode-zig) extension. This has some unfortunate issues:
- some config options are not related to ZLS
- the names don't match since vscode-zig converts them to camelCase
- the default values may be different (e.g `semantic_tokens` is `"partial"` in vscode-zig)

ZLS does provide a [json schema](https://github.com/zigtools/zls/blob/master/schema.json) for all of its configuration options so there is no need to depend on anything else. And it obviously matches exactly what ZLS provides.

### Before
Screenshot from 2024-06-22 13-15-57

### After
Screenshot from 2024-06-22 13-16-39

### DIFF
--- a/packages/zls/package.yaml
+++ b/packages/zls/package.yaml
@@ -84,7 +84,7 @@ source:
 
 
 schemas:
-  lsp: vscode:https://raw.githubusercontent.com/ziglang/vscode-zig/master/package.json
+  lsp: https://raw.githubusercontent.com/zigtools/zls/{{version}}/schema.json
 
 bin:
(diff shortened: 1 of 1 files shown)


williamboman (approved):
Thanks!

PR #6299, opened 2024-06-28
feat: angularls 1 major version per version on version_overrides

zer09:
## Describe your changes
I having trouble on angulals v18, it will exit with error looking for a file named ngtypecheck.ts and I want to other lts version of angularls and it is not available on the registery, the latest version after v18 is v14. With this pull request I added lts version from v9 to v17

Just for a reference there is an open issue open 4 days ago (Jun 24, 2024)
https://github.com/angular/vscode-ng-language-service/issues/2043

### DIFF
--- a/packages/angular-language-server/package.yaml
+++ b/packages/angular-language-server/package.yaml
@@ -17,40 +17,35 @@ source:
     - typescript@5.1.3
 
   version_overrides:
-    - constraint: semver:<=14.1.0
-      id: pkg:npm/%40angular/language-server@14.1.0
+    - constraint: semver:<=17.3.2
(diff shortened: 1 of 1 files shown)


packages/angular-language-server/package.yaml
```diff
@@ -17,41 +17,36 @@ source:
     - typescript@5.1.3
```
williamboman, line 17: Doesn't this version need upgrading too?

zer09: Updated

williamboman: Thanks!

williamboman (approved):
Thanks!

PR #6338, opened 2024-07-02
bump(sonarlint-language-server): update to v4.7.0

thahnen:


williamboman (approved):
Thanks!

PR #6361, opened 2024-07-04
feat(phpactor): update to v2024.06.30.0 and switch to phar installation

Lctrs:
## Describe your changes
In its latest version, phpactor seems to have achieved windows support. We can switch to install their provided PHAR to reduce installation overhead.


williamboman (approved):
Thanks!

PR #6579, opened 2024-07-23
feat: add superhtml package

EliSauder:
## Describe your changes

Added the superhtml lsp to the list of packages that mason can install.

## Screenshots

Example using mason-lspconfig (PR coming soon)
image

LSP Showint up in :LspInfo
image

Installed through Mason
image
image

Not quite ready for prime time since the LSP itself doesn't fully work yet (that or I've set it up wrong) https://github.com/kristoff-it/superhtml/issues/20.

### DIFF
--- a/packages/superhtml/package.yaml
+++ b/packages/superhtml/package.yaml
@@ -1,0 +1,36 @@
+name: superhtml
+description: HTML Language Server & Templating Language Library
+homepage: https://github.com/kristoff-it/superhtml
+licenses:
+  - MIT
+languages:
+  - HTML
+  - SuperHTML
+categories:
+  - LSP
+  - Formatter
+
+source:
+  id: pkg:github/kristoff-it/superhtml@v0.5.3
+  asset:
+    - target: darwin_arm64
+      file: aarch64-macos.tar.gz
+      bin: aarch64-macos/superhtml
+    - target: linux_arm64
+      file: aarch64-linux.tar.gz
+      bin: aarch64-linux/superhtml
+    - target: win_arm64
+      file: aarch64-windows.zip
+      bin: aarch64-windows/superhtml.exe
+    - target: darwin_x64
+      file: x86_64-macos.tar.gz
+      bin: x86_64-macos/superhtml
+    - target: linux_x64_musl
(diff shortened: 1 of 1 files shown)


EliSauder:
@mason-org/triage this should be ready to go now. superhtml has been upversioned to v0.4.3, fixes have been added and some misconfiguration dealt with in the lspconfig PR so that it now all works.

EliSauder:
Some other updates, the superhtml filetype has been added to the vim and neovim repositories, here is the neovim patch: https://github.com/neovim/neovim/commit/e596b6a18d84aa56b5189bb7921feec58279976c

EliSauder:
@williamboman hello, I do apologize for the ping, but this has been sitting getting stale for a while. I was wondering if there is anything else that I can do to help move this along.

tristan957:
Thanks for your work on this :). Looks like we could use pre-built binaries https://github.com/kristoff-it/superhtml/releases/tag/v0.5.0?

EliSauder:
That is true. Just to confirm, @tristan957 this is the sort of setup that I should emulate to use the pre-built packages right?

https://github.com/mason-org/mason-registry/blob/main/packages/zls/package.yaml

txtsd:
Eagerly awaiting this to land in mason and then lazyvim.

EliSauder:
I have updated the package to 0.5.3 and to use the pre-built packages. I additionally tested on my MacOS system that it can get installed with Mason.

tristan957:
I don't think there is anything blocking it. It just seems like the maintainers of this registry have stepped away, at least for now. I think the Neovim community probably needs to fork the registry, so that we can increase the bus factor. The Mason plugin allows us to point at a different registry.

tristan957:
```
    ◍ superhtml
      ▼ Displaying full log
        The current platform is unsupported.
```

Seeing this when trying to install superhtml on Linux x86_64.

tristan957:
I wonder if Mason is not identifying my system as `linux_x64_musl`. I'm on Fedora. I don't see why the binary in the package wouldn't support my glibc based system.

tristan957:
https://github.com/mason-org/mason-registry/pull/8877

PR #6588, opened 2024-07-23
feat: add ruby-lsp-rails

jpeterburs:
## Describe your changes
Added lsp `ruby-lsp-rails`

https://github.com/williamboman/mason.nvim/issues/1434


jpeterburs:
Was able to install the package, but have still some testing to do; will get back to this later

PR #6776, opened 2024-08-06
feat: add daedalus-language-server

pm4rcin:
## Describe your changes
This PR adds a package for (https://github.com/kirides/DaedalusLanguageServer) used by Gothic and Gothic II games.

### DIFF
--- a/packages/daedalus-language-server/package.yaml
+++ b/packages/daedalus-language-server/package.yaml
@@ -1,0 +1,34 @@
+---
+name: daedalus-language-server
+description: A LanguageServer implementation in Go for the scripting language Daedalus used in Gothic and Gothic II games
+homepage: https://github.com/kirides/DaedalusLanguageServer
+licenses:
+  - MIT
+languages:
+  - Daedalus
+categories:
+  - LSP
+
+source:
+  id: pkg:github/kirides/DaedalusLanguageServer@v1.0.0
+  asset:
+    - target: linux_x64_gnu
+      file: linux_amd64.zip
+      bin: DaedalusLanguageServer
+    - target: linux_arm64_gnu
+      file: linux_arm64.zip
+      bin: DaedalusLanguageServer
+    - target: darwin_x64
+      file: darwin_amd64.zip
+      bin: DaedalusLanguageServer
+    - target: darwin_arm64
+      file: darwin_arm64.zip
+      bin: DaedalusLanguageServer
+    - target: win_x64
+      file: windows_amd64.zip
+      bin: DaedalusLanguageServer.exe
+    - target: win_arm64
+      file: windows_arm64.zip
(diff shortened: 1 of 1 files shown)


packages/daedalus-language-server/package.yaml
```diff
@@ -0,0 +1,34 @@
...
+      file: darwin_arm64.zip
+      bin: DaedalusLanguageServer
+    - target: win_x64
+      file: windows_amd64.zip
+      bin: DaedalusLanguageServer.exe
+    - target: win_arm64
+      file: windows_arm64.zip
+      bin: DaedalusLanguageServer.exe
+bin:
+  daedalus: "{{source.asset.bin}}"
```
mehalter, line 34:
```suggestion
  DaedalusLanguageServer: "{{source.asset.bin}}"
```

packages/daedalus-language-server/package.yaml
```diff
@@ -0,0 +1,34 @@
+---
+name: daedalus-language-server
```
mehalter, line 2:
```suggestion
name: DaedalusLanguageServer
```

mehalter (changes requested):
The name of the package should be the name of the project/binary. The package folder name should also be changed to `DaedalusLanguageServer`

packages/daedalus-language-server/package.yaml
pm4rcin: Done

chrisgrieser:
At the mason registry, we decided to only submit packages that are sufficiently popular, for reasons of maintainability and security. It looks like this language server unfortunately does not fulfill the requirements we decided upon. https://github.com/mason-org/mason-registry/blob/main/CONTRIBUTING.md#requirements

I am sorry that you are only hearing about this now and thus invested time in this PR without it being merged. However, you can use the instructions [described here](https://github.com/mason-org/mason-registry/blob/main/CONTRIBUTING.md#testing) to use a registry of your own, so you can nonetheless benefit from the convenience of using mason to install the package.

pm4rcin:
That's unfortunate to hear since criterias were changed on 1 April from what I can see and a few days before that I've received review. As for popularity IIRC it was used by The Chronicles of Myrtana: Archolos team so a serious project. And even THQ Nordic gave credits in Gothic 1 port for Nintendo Switch so can't get more official. But at the end it's your project  your rules. And whatever decision you make I'll respect that.

chrisgrieser:
Yeah, we were in a phase of organizational restructuring back then—sorry for the disorder that resulted from that. 

One reason we decided on the arguably non-low threshold was simply to keep the registry maintainable, since even with additional maintainers there are a lot of PRs piling up. 

Nonetheless, mason supports  using custom registries specifically so tools that can't make it past the threshold can still be used with mason.

PR #6895, opened 2024-08-16
feat: Add refurb

alsmnn:
Adds a package for [refurb](https://github.com/dosisod/refurb), which is a tool for refurbishing and modernizing Python codebases.

## Describe your changes

I would like to add refurb as a python linter.

I have not created a ticket in advance, because I thought it would be just easier to create the PR, but i can create an issue if this is mandatory.

## Screenshots

mason_linter
refurb

### DIFF
--- a/packages/refurb/package.yaml
+++ b/packages/refurb/package.yaml
@@ -1,0 +1,17 @@
+---
+name: refurb
+description: A tool for refurbishing and modernizing Python codebases.
+homepage: https://github.com/dosisod/refurb
+licenses:
+  - GPL-3.0-only
+languages:
+  - Python
+categories:
+  - Linter
+
+source:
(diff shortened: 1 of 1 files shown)


packages/refurb/package.yaml
```diff
@@ -0,0 +1,17 @@
...
+homepage: https://github.com/dosisod/refurb
+licenses:
+  - GPL-3.0-only
+languages:
+  - Python
+categories:
+  - Linter
+
+source:
+  # renovate:versioning=loose
```
williamboman, line 13: I don't see any reason to override the default versioning strategy here
```suggestion
```

williamboman (approved):

williamboman:
@williambotman /fixup

PR #6987, opened 2024-08-24
feat: add xcode-build-server

lbennett-stacki:
## Describe your changes
Adds xcode-build-server to add to mason Swift packages. For use in packages like xcodebuild.nvim.

This tool allows users to build (and more) xcode projects outside of xcode, like in neovim.

After installing, I can build using the tool and the xcodebuild.nvim package.

Screenshot 2024-08-24 at 18 17 48

### DIFF
--- a/packages/xcode-build-server/package.yaml
+++ b/packages/xcode-build-server/package.yaml
@@ -1,0 +1,18 @@
+---
+name: xcode-build-server
+description: A server to bridge Xcode build system outside of Xcode
+homepage: https://github.com/SolaWing/xcode-build-server
+licenses:
+  - MIT
+languages:
+  - Swift
+categories:
+  - LSP
+
+source:
+  id: pkg:github/SolaWing/xcode-build-server@v1.1.0
+  build:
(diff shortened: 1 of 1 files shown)


packages/xcode-build-server/package.yaml
```diff
@@ -0,0 +1,18 @@
...
+  - MIT
+languages:
+  - Swift
+categories:
+  - LSP
+
+source:
+  id: pkg:github/SolaWing/xcode-build-server@v1.1.0
+  build:
+    run: "echo already built"
```
chrisgrieser, line 15: that's unnecessary, `source.build` is actually not required. 

If the release already contains binaries that are read to use, you have to specify them as `assets`. [See this package for an example.](https://github.com/mason-org/mason-registry/blob/main/packages/actionlint/package.yaml)

chrisgrieser (changes requested):

chrisgrieser:
This PR seems abandoned, closing this in favor of the active PR #9776.

PR #7040, opened 2024-08-29
feat: add snakeskin-lsp

sinelnikov-web:
## Describe your changes
Add snakeskin lsp configuration

## Screenshots
image


williamboman (approved):
Thanks! I renamed the package to `snakeskin-cli` because that's how the underlying package is named (also, it seems to be more than just an LSP server)

PR #7297, opened 2024-09-22
tweak: add other shells bash-language-server supports

unrealapex:
## Describe your changes
Add other shells to bash-language-server languages because it would work for other shell configuration files as well.


Conarius:
This won't enable the language server for other shells inside Neovim, just list it for those shells

unrealapex:
You're correct. I'll fix the wording.

PR #4822, opened 2024-03-06
feat: add nixd

plokid:
a nix language server supported by nix-community
by the way, [rnix-lsp](https://github.com/nix-community/rnix-lsp) has archived

PR #6725, opened 2024-08-02
feat: add nixd

antoineco:
## Describe your changes

Adds a package for [nixd](https://github.com/nix-community/nixd), a Nix language server.

Supersedes / closes #4822, which was seemlingly abandonned by its author.

## Checklist before requesting a review

- [x] I have successfully tested installation of the package.
- [x] I have successfully tested the package after installation.
      

## Screenshots

Screenshot 2024-08-02 at 14 21 04

PR #7509, opened 2024-10-10
feat: add nixd

cmpadden:
## Describe your changes

Adds the `nixd` LSP at the latest version 2.4.0. The previous PR has become stale.

Closes https://github.com/williamboman/mason.nvim/issues/1570

Updates the stale, but already approved PR, https://github.com/mason-org/mason-registry/pull/4822

### DIFF
--- a/packages/nixd/package.yaml
+++ b/packages/nixd/package.yaml
@@ -1,0 +1,26 @@
+---
+name: nixd
+description: Language Server for Nix.
+homepage: https://github.com/nix-community/nixd
+licenses:
+  - LGPL-3.0
+languages:
+  - Nix
+categories:
+  - LSP
+
+source:
+  # renovate:datasource=github-tags
+  id: pkg:github/nix-community/nixd@2.4.0
+  build:
+    - target: unix
+      run: |
+        meson -Dprefix="$PWD" build
+        ninja -C build install
+    - target: win
+      run: |
+        meson -Dprefix="($pwd).path" build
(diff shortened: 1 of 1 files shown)


antoineco:
I attempted this in
- https://github.com/mason-org/mason-registry/pull/6725

but I'm not positive that `nixd` can be installed via mason that easily, for three reasons:
- it doesn't offer (pre-built) binary distributions
- it has exotic dependencies which users would need to pre-install for the build to succeed outside of `nix build`
- the build produces shared libraries which `nixd` links against, this isn't supported by Mason

See https://github.com/mason-org/mason-registry/pull/6725#issuecomment-2351015814 for more details.

cmpadden:
> I attempted this in
> 
>     * [feat: add nixd #6725](https://github.com/mason-org/mason-registry/pull/6725)
> 
> 
> but I'm not positive that `nixd` can be installed via mason that easily, for three reasons:
> 
>     * it doesn't offer (pre-built) binary distributions
> 
>     * it has exotic dependencies which users would need to pre-install for the build to succeed outside of `nix build`
> 
>     * the build produces shared libraries which `nixd` links against, this isn't supported by Mason
> 
> 
> See [#6725 (comment)](https://github.com/mason-org/mason-registry/pull/6725#issuecomment-2351015814) for more details.

Gotcha, thanks for outlining that, @antoineco.

I will close this one as well, and share the context in the original GitHub issue too. 🙏

PR #7535, opened 2024-10-13
Install terraform-ls on FreeBSD

gwarf:
Hey,
I'm using Mason via LazyVim on macOS, GNU/Linux and FreeBSD 14.
Various language servers install and works just fine, but for a few I get errors like `The current platform is unsupported.`.
I'm interested in trying to fix the ones that can be fixed, and I'm looking at starting this with terraform-ls.

It turns out that for this one it should in theory be doable and fairly easy as Hashicorp provides packages for terraform-ls on FreeBSD, as an example: https://releases.hashicorp.com/terraform-ls/0.34.3/

I was able to install it by using this repository locally and manually specifying the target:

```shell
:MasonInstall --target=freebsd_x64 terraform-ls
```

Apparently the target is not automatically identified/used, and I'm not sure how/where it should be handled, thanks for your work on Mason and I would welcome any advice regarding how to address this!

### DIFF
--- a/packages/terraform-ls/package.yaml
+++ b/packages/terraform-ls/package.yaml
@@ -21,6 +21,10 @@ source:
       files:
         terraform-ls.zip: https://releases.hashicorp.com/terraform-ls/{{ version | strip_prefix "v" }}/terraform-ls_{{ version | strip_prefix "v" }}_darwin_amd64.zip
       bin: terraform-ls
+    - target: freebsd_x64
+      files:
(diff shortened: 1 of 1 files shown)


gwarf:
So maybe the first thing would be to have FreeBSD be fully supported in Mason and considered as part of https://github.com/williamboman/mason.nvim/blob/main/lua/mason-core/platform.lua?

chrisgrieser:
the registry-lint test is failing and thus blocking the PR since `freebsd_x64` is not included in the schema. So yeah, the schema as well as mason platform support need to be adapter before this PR can go through.

PR #7632, opened 2024-10-21
clangd and yq x86_64 and android compatibility support.

OurCodeBase:
## Changes
- `clangd` server's target is changed from `linux_x64_gnu` to `linux_x64` to support on `x86_64` and android based PCs.
- `yq` server's targets are changed from `linux_x64_gnu` and `linux_arm64_gnu` to `linux_x64` and `linux_arm64` to support on `x86_64` and android based PCs.
- `yq` server's `gnu` linux distro targets are shifted under `openbsd` targets to kill conflicts.

### DIFF
--- a/packages/clangd/package.yaml
+++ b/packages/clangd/package.yaml
@@ -18,7 +18,7 @@ source:
     - target: [darwin_x64, darwin_arm64]
       file: clangd-mac-{{version}}.zip
       bin: clangd_{{version}}/bin/clangd
-    - target: linux_x64_gnu
+    - target: linux_x64
       file: clangd-linux-{{version}}.zip
       bin: clangd_{{version}}/bin/clangd
     - target: win_x64
--- a/packages/yq/package.yaml
+++ b/packages/yq/package.yaml
@@ -17,12 +17,6 @@ source:
     - target: darwin_x64
       file: yq_darwin_amd64.tar.gz
       bin: yq_darwin_amd64
-    - target: linux_x64_gnu
-      file: yq_linux_amd64.tar.gz
-      bin: yq_linux_amd64
-    - target: linux_arm64_gnu
-      file: yq_linux_arm64.tar.gz
-      bin: yq_linux_arm64
     - target: linux_x64_openbsd
       file: yq_openbsd_amd64.tar.gz
       bin: yq_openbsd_amd64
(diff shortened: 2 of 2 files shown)


chrisgrieser (commented):
Sorry for the late response. There is an issue with the linux_arm64 for `yq` that is blocking the merge.

williamboman:
These have a `_gnu` suffix because the compiled binaries are dynamically linked using the GNU linker. "GNU support" is [determined by the mason.nvim client](https://github.com/mason-org/mason.nvim/blob/main/lua/mason-core/platform.lua#L51-L65). If you're on a system that supports executing GNU binaries as-is (i.e. no nix-ld or something like that) but it's not getting recognized by Mason I'd be more interested in addressing that in `mason.nvim` than modifying the package like this.

OurCodeBase:
> These have a `_gnu` suffix because the compiled binaries are dynamically linked using the GNU linker. "GNU support" is [determined by the mason.nvim client](https://github.com/mason-org/mason.nvim/blob/main/lua/mason-core/platform.lua#L51-L65). If you're on a system that supports executing GNU binaries as-is (i.e. no nix-ld or something like that) but it's not getting recognized by Mason I'd be more interested in addressing that in `mason.nvim` than modifying the package like this.

Thanks for the response,

I'm using debian for development and sometimes i use termux on my android phone. I've tested this and it's working fine on both of them.

Can I use the forked version by default?

robertkirkman:
Hi, I have created a port of [`codelldb`](https://github.com/vadimcn/codelldb) to Android that is compiled and linked with the Android bionic C library and linker.

I've deployed a first revision of `codelldb` for Android into the Termux package repositories, but I am also working on a second revision of it here that has several things fixed:

- https://github.com/termux/termux-packages/pull/25663

I noticed that in this repository, there's an instance of a `codelldb` package,

https://github.com/mason-org/mason-registry/blob/9c8dd7818ac5ba089b4920945cb23ed4ac2b8493/packages/codelldb/package.yaml#L4

however, unfortunately this repository doesn't support Android, and OpenVSIX doesn't support Android either, so I am confused about the best way to proceed with distributing the functioning build of the Rust code of `codelldb` into an Android Bionic C library executable that I have created.

I am pretty sure that I will **probably be able to distribute all the necessary packages for `codelldb` on Android through the Termux hosting infrastructure,**

I'm just explaining the situation that I'm in here so that it's clear what I have been doing regarding the situation of using a different sort of "editor plugin" that is also here, on Android.

I was thinking that maybe, if this repository had support for an "Android" target of some kind, it should most likely be "`linux_x64_android`", etc, and "`linux_x64_gnu`" should be left unchanged, since they are relatively cleanly separated operating systems that both have a Linux kernel, especially when Termux is considered (which is the primary distribution host of Neovim for Android and the dependencies of that)?

PR #7642, opened 2024-10-22
feat: add c3-lsp

williamboman:
- **Add C3 LSP**
- **fixes**


Conarius:
There's also a [vscode lsp schema](https://github.com/pherrymason/c3-lsp/blob/main/client%2Fvscode%2Fpackage.json) available for it

PR #7646, opened 2024-10-22
feat(ruby-lsp): add vscode lsp schema

Conarius:
## Describe your changes
Adds vscode lsp schema to ruby-lsp


mehalter (approved):
Thanks for all of these schema additions!

PR #7681, opened 2024-10-25
feat: add cppcheck linter

saifazmi:
## Describe your changes
Hi, I saw that [cppcheck](https://cppcheck.sourceforge.io/) was missing from Mason registry so I wanted to add that in.

It has the following build config:
- default release build
- does not support custom "rules" feature (using regex)
  - rules feature requires external dependency on [PCRE](https://www.pcre.org/) package

Explicitly using `-DFILESDIR` build flag to workaround issue: https://github.com/danmar/cppcheck/pull/6764

- [x] I have successfully tested installation of the package from local `mason-registry`
- [x] I have successfully tested the package after installation using `nvim-lint`

## Screenshots
cppcheck_nvim-lint_test

### DIFF
--- a/packages/cppcheck/package.yaml
+++ b/packages/cppcheck/package.yaml
@@ -17,7 +17,7 @@ source:
   build:
     - target: unix
       run: |
-        NVIM_DATA_PATH=$(nvim --headless -c 'lua print(vim.fn.stdpath("data"))' -c 'q' 2>&1)
+        NVIM_DATA_PATH=$(nvim --headless -c "lua print(vim.fn.stdpath('data'))" -c "q" 2>&1)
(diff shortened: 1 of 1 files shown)


saifazmi:
cc @williamboman all checks are green and I have tested on local that it works as expected. Should be an straight forward merge.

saifazmi:
hi, can anyone help review and merge this, please?
all checks are green, should be an easy merge.

williamboman (commented):
Hey and thanks for the PR! Custom build scripts are being deprecated because of difficulty in maintaining them over time and their common reliance on system dependencies that Mason cannot detect. You can always host your own registry if you'd like to manage it through Mason, see https://github.com/williamboman/mason-registry-playground/ for a starting point

PR #7729, opened 2024-10-30
feat(neocmakelsp): use release assets

eeeXun:
## Describe your changes

Use release assets since neocmakelsp provide the binary file, and removes dependency on cargo

### DIFF
--- a/packages/neocmakelsp/package.yaml
+++ b/packages/neocmakelsp/package.yaml
@@ -18,8 +18,8 @@ source:
       file: neocmakelsp-x86_64-apple-darwin
     - target: linux_x64
       file: neocmakelsp-x86_64-unknown-linux-musl
-    - target: linux_arm64
-      file: neocmakelsp-aarch64-unknown-linux-musl
+    - target: linux_arm64_gnu
+      file: neocmakelsp-aarch64-unknown-linux-gnu
     - target: win_x64
       file: neocmakelsp-x86_64-pc-windows-msvc.exe
 
--- a/packages/neocmakelsp/package.yaml
+++ b/packages/neocmakelsp/package.yaml
@@ -10,7 +10,23 @@ categories:
   - LSP
 
 source:
-  id: pkg:cargo/neocmakelsp@0.8.12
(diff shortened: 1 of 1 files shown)


Conarius:
I think we've had that discussion somewhere else.
But it can have a downside to use release assets:
- You are dependent on the release asset names and also architectures
So if the names change, this change has to be reflected here. And if they decided to drop or add support for a platform, it has to be done here too.
So using cargo makes it a bit longer for the end user to have the binary compiled, but it will work on their machine no matter the release assets.

But ultimately it's up to the maintainer to approve or deny it.

eeeXun:
> I think we've had that discussion somewhere else. But it can have a downside to use release assets:

Would you mind giving me the link to the discussion. I want to see the conclusion of discussion.

> * You are dependent on the release asset names and also architectures
>   So if the names change, this change has to be reflected here. And if they decided to drop or add support for a platform, it has to be done here too.
>   So using cargo makes it a bit longer for the end user to have the binary compiled, but it will work on their machine no matter the release assets.

That's absolutely true. The changes will increase the burden of maintainer.

IMHO, as this project is used by many people, these changes not only save the network bandwidth, but also save the power consumption when compiling binary files.

Conarius:
> Would you mind giving me the link to the discussion. I want to see the conclusion of the discussion.
https://github.com/mason-org/mason-registry/pull/6057

> That's absolutely true. The changes will increase the burden on maintainer.
> 
> IMHO, as this project is used by many people, these changes not only save the network bandwidth, but also save the power consumption when compiling binary files.

Yeah, you can make that argument but I do think most end users don't really care about these that much. They most likely care about it taking a long time to update and such.

But yeah. It's ultimately up to the maintainer to make the decision

eeeXun:
Okay, I didn't notice there's already a PR. Close this PR and track here.

Decodetalkers:
Can mason use fallback way to download asserts when the target platform asserts are not provided?

williamboman:
Prebuilt assets are preferred as long as it supports at least linux_x64 and darwin_arm64 :+1:

PR #7807, opened 2024-11-05
feat(remark-ls): add vscode lsp schema

Conarius:
## Describe your changes
Adds vscode lsp schema to remark ls. The versions between the vscode repo and the language server are identical though only a few days apart with the vscode extenstion being the first.


mehalter (changes requested):
The schema is in a separate project with versioning that does not always match up. This might not be a stable way to link the schema to the correct language server version

PR #7863, opened 2024-11-07
feat(custom-elements-ls): add vscode lsp schema

Conarius:
## Describe your changes
Adds vscode lsp schema to custom-elements-languageserver

### DIFF
--- a/packages/custom-elements-languageserver/package.yaml
+++ b/packages/custom-elements-languageserver/package.yaml
@@ -12,5 +12,8 @@ categories:
 source:
   id: pkg:npm/custom-elements-languageserver@1.0.4
 
+schemas:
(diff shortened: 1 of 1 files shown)


chrisgrieser:
@Conarius As there are more than 50 PRs from you adding LSP schemas, wouldn't it maybe make more sense so batch them into one big PR? Considering that there are 150+ open PRs for the registry, this would also significantly reduce the maintenance burden for @williamboman when he has time to merge the PRs.

(Btw, what an endeavor to open 50 PRs! That's some commendable effort.)

Conarius:
@chrisgrieser True. Though it probably has to be tested and some might be discarded. Having it one bundled pr might be tricky then when every one of them will be tested and some will just be closed

mehalter (approved):

PR #8037, opened 2024-11-26
feat: add gdtoolkit3

ignoxx:
## Describe your changes
This PR adds a `gdtoolkit3` package which specifically is intendend to use for Godot 3 only. The already existing [`gdtoolkit`](https://github.com/mason-org/mason-registry/tree/main/packages/gdtoolkit) package only delivers tools for Godot 4 which do not work with Godot 3

I coudn't figure out if i can somehow reuse the same `gdtoolkit` package and define both versions in it without pretty much just duplicating the existing one :)

### DIFF
--- a/packages/gdtoolkit3/package.yaml
+++ b/packages/gdtoolkit3/package.yaml
@@ -1,0 +1,18 @@
+---
+name: gdtoolkit3
+description: A set of tools for daily work with GDScript for Godot 3.
+homepage: https://github.com/Scony/godot-gdscript-toolkit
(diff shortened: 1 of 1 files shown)


Conarius:
It's just a version constraint as far as I can see. You can install specific versions with mason already, so this makes this pr very much obsolete

ignoxx:
> It's just a version constraint as far as I can see. You can install specific versions with mason already, so this makes this pr very much obsolete

oh great! obsolete indeed, thx

PR #8084, opened 2024-12-01
add-flake8-pyproject

mitchdost:
## Describe your changes
Request for providing [john-hen/Flake8-pyproject](https://github.com/john-hen/Flake8-pyproject) as a Mason package.

I was not shure about where this information to put, so I added it as second line of the description:

> Mason: To be used as a replacement for package flake8, because it also provides the flake8 binary.


mehalter:
@mason-org/triage this might take some discussion since it provides the same binary as just the `flake8` package. Not sure if there is a way in the Mason registry to set constraints on packages such as not having 2 packages installed at the same time.

PR #8182, opened 2024-12-10
feat: add veridian

Conarius:


mehalter:
Not sure what to do here in terms of versioning before there is a stable versioning setup for the package. Looks like they are releasing a "nightly" but it wouldn't update if the commit changes for that version tag.

Conarius:
@mehalter 
Only other option would be building from source for every commit and using commits as the refs

chrisgrieser (approved):
I'd say, merge it for now, and change this package spec once the package has more stable releases?

PR #8216, opened 2024-12-13
feat(golang): add gosrt formatter

akarl:
## Describe your changes
gosrt sorts top-level symbols such as functions and types in the same order as godoc.

### DIFF
--- a/packages/gosrt/package.yaml
+++ b/packages/gosrt/package.yaml
@@ -1,0 +1,16 @@
+---
+name: gosrt
+description: A Go formatter that sorts top-level symbols like godoc.
+homepage: https://gitlab.com/biffen/gosrt
+licenses:
+  - MIT
+languages:
+  - Go
+categories:
+  - Formatter
+
+source:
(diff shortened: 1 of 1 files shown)


akarl:
It would be cool to merge this :)

akarl:
Still relevant!

chrisgrieser:
Sorry for the late response!

At the Mason registry, we decided to only submit packages that are sufficiently popular, for reasons of maintainability and security. It looks like this package unfortunately does not fulfill the [requirements for new packages](https://github.com/mason-org/mason-registry/blob/main/CONTRIBUTING.md#requirements) we decided upon. 

(GitLab stars are not really meaningful, since few people have a GitLab account. But I also have not found any other references to the project that are less than 5 years old, and the repo itself seems to not have received any updates in the last 2 years.)

I am sorry that we are just communicating this now, and you have thus invested time in this PR without it being merged. However, you can use the instructions [described here](https://github.com/mason-org/mason-registry/blob/main/CONTRIBUTING.md#testing) to use a registry of your own, so you can nonetheless benefit from the convenience of using Mason to install the package.

Issue #8015, opened 2024-11-23
feat: add typstyle formatter


UtkarshVerma:
`typstfmt` which is present in the registry has been archived and the author recommends using `typstyle` instead.

https://github.com/Enter-tainer/typstyle/

PR #8223, opened 2024-12-13
fix: Remove unmaintained typstfmt

lvgrr:
## Describe your changes
The [repository](https://github.com/astrale-sharp/typstfmt) for `typstfmt` has been archived and the project is currently unmaintained.

This PR deletes `typstfmt` from the registry.

Relates to #8015.

- [x] I have successfully tested deletion of the package.


Conarius:
A deprecation note would suffice I think

chrisgrieser:
I agree. Using the deprecation status not only creates less surprises for the user (why is a tool suddenly gone?), but also does not result in some mason-related surroundings to behave weirdly (like auto-uninstalling the package).

chrisgrieser (changes requested):
please use the `deprecated` status instead of removing the package

PR #8122, opened 2024-12-04
bump(sonarlint-language-server): update to v4.14.0

coderkun:
Update SonarLint language server to version [4.14](https://github.com/SonarSource/sonarlint-vscode/releases/tag/4.14.0%2B77092).

PR #8251, opened 2024-12-17
bump(sonarlint-language-server): update to v4.14.0

thahnen:


williamboman:
Thanks, however this was merged already in https://github.com/mason-org/mason-registry/pull/8122!

PR #8350, opened 2024-12-26
Add package for marko-language-server

russell-knight:
## Describe your changes

Add language server for Marko-js

## Screenshots

image
image
image

### DIFF
--- a/packages/marko-js/package.yaml
+++ b/packages/marko-js/package.yaml
@@ -1,0 +1,12 @@
+name: marko-language-server
+description: Language Server for Marko
+homepage: https://github.com/marko-js/language-server
+licenses:
+  - MIT
+languages:
+  - Marko
+categories:
+  - LSP
+
+source:
+    id: pkg:npm/%40marko/language-server@1.4.7


packages/marko-js/package.yaml
```diff
@@ -0,0 +1,12 @@
...
+homepage: https://github.com/marko-js/language-server
+licenses:
+  - MIT
+languages:
+  - Marko
+categories:
+  - LSP
+
+source:
+    id: pkg:npm/%40marko/language-server@1.4.7
```
mehalter, line 12:
```suggestion
    id: pkg:npm/@marko/language-server@1.4.7
    
schemas:
  lsp: vscode:https://raw.githubusercontent.com/marko-js/language-server/@marko/language-server@{{version}}/packages/vscode/package.json
  
bin:
  marko-language-server: npm:@marko/language-server
```

We should include the LSP schema with the package specification as well as the binary

mehalter (changes requested):
The package folder name should also be changed to `marko-language-sever` and not `marko-js`

packages/marko-js/package.yaml
Conarius: > We should include the LSP schema with the package specification as well as the binary

True, but the %40 can stay, has to do with encoding

Conarius:
```suggestion
    id: pkg:npm/%40marko/language-server@1.4.7
    
schemas:
  lsp: vscode:https://raw.githubusercontent.com/marko-js/language-server/@marko/language-server@{{version}}/packages/vscode/package.json
  
bin:
  marko-language-server: npm:%40marko/language-server
```

Like this

Conarius (changes requested):

packages/marko-js/package.yaml
mehalter: Yeah the main thing is that the `%40` is not necessary here. There are many other packages that just use `@` in the `npm` string and they seem to work. So it seems better to use `@` for readability

russell-knight: Thanks both for the review. It's been a while since I first opened this PR so it's hard to remember clearly, but I believe I was having trouble getting this working without the `%40` encoding which is why I included it. I'm on a Windows machine so it might have been something to do with that? Either way I've updated this PR to:
- change the directory to `marko-language-server`
- add the schema and bin to the config file

Conarius (approved):
LGTM

mehalter (approved):

PR #7595, opened 2024-10-18
feat: add tex-fmt

marcfranquesa:
## Describe your changes
Add tex-fmt formatter for LaTeX files.

## Checklist before requesting a review
- [X] I have successfully tested installation of the package.
- [X] I have successfully tested the package after installation.

## Screenshots

PR #8382, opened 2024-12-30
tex-fmt: add package

WGUNDERWOOD:

### DIFF
--- a/packages/tex-fmt/package.yaml
+++ b/packages/tex-fmt/package.yaml
@@ -1,0 +1,16 @@
+---
+name: tex-fmt
+description: LaTeX formatter written in Rust
+homepage: https://github.com/WGUNDERWOOD/tex-fmt
+licenses:
+  - MIT
+languages:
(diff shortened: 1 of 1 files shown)


ficcdaf:
I think the Cargo source here is better than the binary tarball source. I eagerly await this package being in Mason.

WGUNDERWOOD:
> I think the Cargo source here is better than the binary tarball source. I eagerly await this package being in Mason.

I thought I had specified the source as being from cargo; do I need to change anything?

ficcdaf:
> > I think the Cargo source here is better than the binary tarball source. I eagerly await this package being in Mason.
> 
> I thought I had specified the source as being from cargo; do I need to change anything?

nope sorry I didn't communicate that clearly. I meant that I like what you did here. There was another PR for adding this package to Mason that used the binary tarballs as sources instead; I think the `cargo` approach is better.

williamboman:
Closing in favor of #7595

PR #8617, opened 2025-01-19
Update `nimlangserver` package name for semver==1.8.0

HexSegfaultCat:
## Describe your changes
Nim's [`nimlangserver`](https://github.com/nim-lang/langserver/releases) LSP in latest version (1.8.0) removed the version from its tar.gz file name. I removed that for versions >= 1.8.0. I tested both 1.8.0 and 1.6.0 versions and they seem to work (installation and the LSP itself).


williamboman (commented):
Thanks for the PR! I managed to adjust this in https://github.com/mason-org/mason-registry/pull/8578 before I saw this, so closing this here

PR #8813, opened 2025-02-05
feat: add emmylua_ls

mawkler:

It hasn't been added to nvim-lspconfig yet

### DIFF
--- a/packages/emmylua_ls/package.yaml
+++ b/packages/emmylua_ls/package.yaml
@@ -1,0 +1,16 @@
+---
+name: emmylua_ls
+description: The language server for Lua, offering extensive features for different Lua versions.
+homepage: https://github.com/CppCXY/emmylua-analyzer-rust/tree/main/crates/emmylua_ls
+licenses:
+  - MIT
+languages:
+  - Lua
+categories:
+  - LSP
+
+source:
+  id: pkg:cargo/emmylua_ls@0.4.4
+
+bin:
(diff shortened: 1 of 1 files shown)


mehalter:
@mason-org/triage do we have a preference for using external package managers vs published binaries from the github releases? It looks like this project does have binaries to be downloaded rather than installing from Cargo

chrisgrieser:
> @mason-org/triage do we have a preference for using external package managers vs published binaries from the github releases? It looks like this project does have binaries to be downloaded rather than installing from Cargo

Imo, GitHub releases are the most preferable option, since any package managers effectively adds a dependency that the user has to have installed before being able to use mason to install this package. (Though they do have the drawback that a change of release asset names requires a PR to fix them). Very common package managers, such as `npm` or `pypi`, are also fine, since we can assume that the majority of people have it installed. 

But I don't think the majority of people have the rust tool chain (`cargo`) installed, so GitHub releases should definitely be used if they are available.

mawkler:
I've changed to using releases now instead.

I rebased on main, and I'm not sure why GitHub suddenly thinks that I'm adding 700 commits 🤔

mawkler:
@mehalter Thank you! What was the issue?

mehalter:
idk the git history was just all messed up. So I cherry picked the 2 commits onto a new branch from main and then pushed that to your branch

chrisgrieser (approved):

mehalter (approved):
This looks good to me!

PR #8877, opened 2025-02-10
Add superhtml support for linux_x64_gnu

tristan957:


EliSauder:
Assuming we want to keep using binaries, this should be opened on the superhtml repository. However, building from source is also an option in the setup.

tristan957:
> Assuming we want to keep using binaries, this should be opened on the superhtml repository. However, building from source is also an option in the setup.

What should be opened on their repository?

EliSauder:
> What should be opened on their repository?

Currently the superhtml repo only provide a musl build for linux. If gnu can use that, then we can just use it for both (I don't know if that works or not).

williamboman:
We can actually condense this into a single `linux_x64` entry since it's a statically compiled binary. I didn't catch the `linux_x64_musl` in the original PR, this target is only needed when the binary is a dynamically linked musl binary (very rare).

EliSauder:
> We can actually condense this into a single `linux_x64` entry since it's a statically compiled binary. I didn't catch the `linux_x64_musl` in the original PR, this target is only needed when the binary is a dynamically linked musl binary (very rare).

Ah, that is good to know.

PR #9018, opened 2025-02-20
feat(bash-debug-adapter): expose `bashdb` as a shared directory

mehalter:
## Describe your changes

The `bashdb` directory is important for setting up the debugger with `nvim-dap` and should be exposed with the `share/` directory. This is part of the efforts to help with the migration to Mason v2


mehalter (commented):
Just updated this with the same change requested here: https://github.com/mason-org/mason-registry/pull/9019#pullrequestreview-2711920151

I believe it also applies. Ready for review from other eyes

Conarius (approved):
LGTM

PR #9019, opened 2025-02-20
feat(codelldb): expose the `lldb` library as a shared folder

mehalter:
## Describe your changes
The `lldb` folder with the binaries and the library files are important files for setting up the debugger and should be marked as stable files. This is part of the efforts for the migration to Mason v2

### DIFF
--- a/packages/codelldb/package.yaml
+++ b/packages/codelldb/package.yaml
@@ -59,3 +59,6 @@ source:
 
 bin:
   codelldb: "{{source.asset.bin}}"
+
+opt:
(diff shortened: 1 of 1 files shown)


williamboman (commented):
Nit: I believe this directory contains binary files for the given architecture? The `share` directory is primarily intended for architecture independent files. I think `opt:` might be a better location (`$MASON/opt/lldb`)

mehalter (commented):
I think this is good to go now

Conarius (approved):
LGTM

PR #9053, opened 2025-02-23
feat: Add local-lua-debugger-vscode

mixam24:
## Describe your changes
Add `local-lua-debugger-vscode`

### DIFF
--- a/packages/local-lua-debugger-vscode/package.yaml
+++ b/packages/local-lua-debugger-vscode/package.yaml
@@ -1,0 +1,17 @@
+---
+name: local-lua-debugger-vscode
+description: A simple Lua debugger which requires no additional dependencies.
+homepage: https://github.com/tomblind/local-lua-debugger-vscode
+licenses:
+ - MIT
+languages:
+ - Lua
+categories:
+ - DAP
+source:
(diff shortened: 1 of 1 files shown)


mehalter (approved):
This looks good to me!

chrisgrieser (approved):
From what I can tell, this looks good (though I have to say I haven't seen a package using `share` before, and after reading the spec, I am not fully sure how mason treats this differently from `bin`, since both seem to be symlinked?)

mehalter:
@chrisgrieser there are some important stuff with share and opt folders that Mason supports. It's basically for non binary files that are shipped with packages where we define a stable file location state. Files in the package installation paths are considered unstable and the share/opt folders (along with bin) provide what the user should be considered as stable paths.

Some packages might only install a library rather than a binary that is useful for the user which is where share comes in

mehalter:
A few packages use this notation and it's becoming a lot more important in Mason v2 where it will be a much stronger requirement to not use paths into the package installation folder.

There are a couple other PRs open to help with this:

- https://github.com/mason-org/mason-registry/pull/9019
- https://github.com/mason-org/mason-registry/pull/9018

PR #9313, opened 2025-03-17
feat: add qmlls

deeedob:
> The qmlls is now provided as prebuilt binary.
> Nightly versions currently follow an incompatible version scheme and will be fixed upstream.

What is the recommended strategy to update it here in the future? :)

### DIFF
--- a/packages/qmlls/package.yaml
+++ b/packages/qmlls/package.yaml
@@ -1,0 +1,28 @@
+---
+name: qmlls
+description: |
+  QML Language Server is a tool shipped with Qt that helps you write code in your favorite (LSP-supporting) editor.
+homepage: https://doc-snapshots.qt.io/qt6-dev/qtqml-tooling-qmlls.html
+licenses:
+  - LicenseRef-Qt-Commercial
+  - GPL-3.0-only WITH Qt-GPL-exception-1.0
+languages:
+  - QML
+categories:
+  - LSP
+
+source:
+  id: pkg:github/TheQtCompanyRnD/qmlls-workflow@0.2
+  asset:
+    - target: [darwin_x64, darwin_arm64]
+      file: qmlls-macos-{{version}}.zip
+      bin: qmlls
+    - target: linux_x64_gnu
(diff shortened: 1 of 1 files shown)


mehalter:
Not sure what to do with this. The licenses are open source (the source lives here: https://code.qt.io/cgit/qt/qtdeclarative.git/tree/) and the distribution repository that we get the binary from is also official from QT. The problem is the licenses are custom and do not have official SPDX strings.

deeedob:
> that we get the binary from is also official from QT

I've not seen any qmlls binaries in the neovim ecosystem. Where are they used and where do you get the binaries from? We (at Qt) just started to provide official binaries to the statically linked qmlls. This is the most up-to-date version and officially supported. It's also used in the Qt vscode extension.

> The problem is the licenses

Whatever binaries have been used follow the exact same licensing. It's `LicenseRef-Qt-Commercial OR GPL-3.0-only WITH Qt-GPL-exception-1.0.` -> https://contribute.qt-project.org/quips/18#tools-and-utils

deeedob:
btw. nightly version mismatch for the tags has been fixed upstream :)

mehalter:
Yeah I see what you mean! My comments were more for the other maintainers here as to how we should proceed. Thanks for the information!

deeedob:
Currently there are also no licenses included in the download, but we will probably add them. Perhaps this will resolve the issue with custom SPDXs -- not sure though, this is not my specialty

With recent changes in qmlls, I also plan to update nvim-lspconfig once this patch is in :)

Let me know what's missing in this PR and I'll update it!

deeedob:
ping, how can we proceed or what's holding this back?

williamboman (approved):
lgtm

deeedob:
Nice! Thanks for the work. How does updating work now? Manually or is there a bot which tracks release tags?

deeedob:
Ping, should we merge it :D ?

Conarius (approved):
LGTM

PR #9363, opened 2025-03-20
feat: add plantuml-lsp

ptdewey:
## Describe your changes
Add plantuml-lsp

### DIFF
--- a/packages/plantuml-lsp/package.yaml
+++ b/packages/plantuml-lsp/package.yaml
@@ -10,9 +10,10 @@ categories:
   - LSP
 
 source:
-  id: pkg:github/ptdewey/plantuml-lsp@v0.4.0
-  build:
-    run: go build
+  id: pkg:golang/github.com/ptdewey/plantuml-lsp@v0.4.0
+
+schemas:
+  lsp: vscode:https://raw.githubusercontent.com/ptdewey/planuml-lsp/{{version}}/editors/code/package.json
 
 bin:
   plantuml-lsp: golang:plantuml-lsp
--- a/packages/plantuml-lsp/package.yaml
+++ b/packages/plantuml-lsp/package.yaml
@@ -1,0 +1,18 @@
+---
+name: plantuml-lsp
+description: Language server for PlantUML.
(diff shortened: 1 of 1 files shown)


packages/plantuml-lsp/package.yaml
```diff
@@ -0,0 +1,18 @@
...
+  - MIT
+languages:
+  - PlantUML
+categories:
+  - LSP
+
+source:
+  id: pkg:github/ptdewey/plantuml-lsp@v0.4.0
+  build:
+    run: go build
```
mehalter, line 15:
```suggestion
  id: pkg:golang/github.com/ptdewey/plantuml-lsp@v0.4.0

schemas:
  lsp: vscode:https://raw.githubusercontent.com/ptdewey/planuml-lsp/{{version}}/editors/code/package.json
```

We should use the `golang` installer rather than manually calling `go build`

This also recommends adding in the VS Code extension schema. This won't work until it is merged in (I see you have an open PR) but is for the future.

ptdewey:
>  We may need to wait until the project gains some more traction and gets more stars (goal is 100+) or when the project is merged into `nvim-lspconfig`

This sounds reasonable to me, I didn't know about the lspconfig 100 star requirement until after I made this PR, so that is on me. It makes sense not to merge until I meet the requirements for merging there also.

mehalter:
No worries at all! The requirements were a decision we have made today actually 😅  We are putting together a new maintainer team to help get these PRs processed more quickly and help with the longevity of the project and are discussing lots of things to make sure we can maintain the project well.

packages/plantuml-lsp/package.yaml
ptdewey: Oh that does seem nice, I might have to check over my package definition to ensure everything is well documented there (the VSCode extension was kind of an afterthought that I haven't put a lot of time into thus far).

I didn't see anything about the `schemas` option in the documentation, maybe it would be worth adding there somewhere?

ptdewey:
I was talking about the requirements for merging into `nvim-lspconfig` there, I hadn't checked them until after I made this PR, and my project definitely doesn't meet those requirements yet.

PR #9372, opened 2025-03-21
Added robotcode LSP and DAP package

kimfaint:
## Describe your changes
Added a package for the [robotcode ](https://robotcode.io/) language server and debugger, used with [Robot Framework](https://robotframework.org/).

Note that the other package robotframework_lsp from robocorp is [no longer being maintained](https://github.com/robocorp/robotframework-lsp/issues/1051) and [robotcode has taken over as the tool of choice for vscode](https://docs.robotframework.org/docs/getting_started/ide#visual-studio-code).

## Screenshots

image

image

### DIFF
--- a/packages/robotcode/package.yaml
+++ b/packages/robotcode/package.yaml
@@ -1,0 +1,17 @@
+---
+name: robotcode
+description: The Ultimate Robot Framework Toolset
+homepage: https://robotcode.io/
+licenses:
+  - Apache-2.0
+languages:
+  - Robot Framework
+categories:
+  - LSP
+  - DAP
+
+source:
+  id: pkg:pypi/robotcode[all]@1.0.3
+
+bin:
+  robotcode: pypi:robotcode
--- a/packages/robotcode/package.yaml
+++ b/packages/robotcode/package.yaml
@@ -13,5 +13,9 @@ categories:
 source:
   id: pkg:pypi/robotcode[all]@1.0.3
 
+
+schemas:
+  lsp: vscode:https://raw.githubusercontent.com/robotcodedev/robotcode/v{{version}}/package.json
+
 bin:
   robotcode: pypi:robotcode


d-biehl:
Thanks @kimfaint for this PR!!!

However, there was already another attempt to get RobotCode into the Mason registry — see https://github.com/mason-org/mason-registry/pull/4886. The installation worked, but the RobotCode Language Server ran in its own Python virtual environment. RobotCode actually needs to run in the same Python environment where tests are executed (i.e., where the libraries you’re using and Robot Framework are installed), otherwise loading libraries, resources, variables — and even debugging — won’t work. That earlier PR was ultimately abandoned due to time constraints and gaps in knowledge.

I think it might be better to use a wrapper script that locates RobotCode in the correct virtual environment (and installs it if necessary) and then launches it there. That seems like the proper approach — could you look into whether and how this can be done?

By the way, there’s already an issue in the RobotCode repo about Vim support (https://github.com/robotcodedev/robotcode/issues/232). I’ve mentioned this PR in that issue; we can discuss the technical details there.

kimfaint:
Thanks @d-biehl, I only tried with local libraries, variables and resources, by setting them in my `PYTHONPATH`, but I see your point with handling external libraries, this won't work.  I'll move my focus to robotcodedev/robotcode#232

packages/robotcode/package.yaml
```diff
@@ -0,0 +1,17 @@
...
+licenses:
+  - Apache-2.0
+languages:
+  - Robot Framework
+categories:
+  - LSP
+  - DAP
+
+source:
+  id: pkg:pypi/robotcode[all]@1.0.3
```
mehalter, line 14:
```suggestion
  id: pkg:pypi/robotcode[all]@1.0.3


schemas:
  lsp: vscode:https://raw.githubusercontent.com/robotcodedev/robotcode/v{{version}}/package.json
```

We should add the schema for the LSP configuration

mehalter (changes requested):

mehalter (approved):
This looks great. The issues indicated in the previous attempt at this are regarding proper configuration of the language server in Neovim and properly setting up the language server settings with the proper python path. This seems to be good to merge.

mehalter:
We just need a 2nd set of eyes and we can get this merged in

chrisgrieser (approved):

chrisgrieser:
@jchabrol 
generally, problems with configuring an LSP should be submitted to the LSP's repo. 

If it is related to the mason registry entry of the LSP, then you should submit a bug report at the mason registry – comments at PRs tend to have low visibility.

ThangEthan:
I can't get to work on my lazyvim. The language server doesn't start.

Conarius:
> I can't get to work on my lazyvim. The language server doesn't start.

@ThangEthan 
What is your setup for robotcode with LazyVim? It should be relatively easy now with vim.lsp.config und vim.lsp.enable since 0.10 and 0.11

chrisgrieser:
@ThangEthan I really don't want to sound rude, but literally the comment above you mentions that in case of a bug, you should submit an issue instead of making a comment here. 

---

Do also note that if this is about lazyvim, it might also be an issue related to lazyvim and not mason.

kimfaint:
@ThangEthan any logs you can share?  I'm not familiar with lazyvim.  With neovim a log is written to `~/.local/state/nvim/lsp.log`.

Another thing to check is that you have the langageserver extra installed for robotcode in your virtual environment, e.g. `pip install robotcode[langageserver]` or for all of the extras `pip install robotcode[all]`

I'm using [nvim-cmp](https://github.com/hrsh7th/nvim-cmp) for code completion and I found I need this additional nvim config for it to run the correct python in my venv:

```lua
local pythonpath = os.getenv("PYTHONPATH")
local handle = io.popen('which python')
local output = handle:read('*a')
local format = output:gsub('[\n\r]', '')
local pythonexe = format
print("pythonexe:", pythonexe)

require('lspconfig')['robotcode'].setup {
    on_attach = my_on_attach,
    capabilities = capabilities,
    settings = {
        robot = {
            pythonpath = {
                pythonpath
            },
            python = {
                executable = pythonexe
            }
        }
    }
}
```

But as mentioned, none of this is a mason issue.

DudeNr33:
For LazyVim, I used this configuration in `~/.config/nvim/lua/plugins/robot.lua`:

```lua
return {
  "neovim/nvim-lspconfig",
  opts = {
    servers = {
      robotcode = {
        -- default fallback
        cmd = { "robotcode", "language-server" },

        -- dynamically adjust cmd depending on lockfiles in root_dir
        on_new_config = function(new_config, root_dir)
          local uv = vim.loop

          if uv.fs_stat(root_dir .. "/uv.lock") then
            new_config.cmd = { "uv", "run", "robotcode", "language-server" }
            return
          end

          if uv.fs_stat(root_dir .. "/poetry.lock") then
            new_config.cmd = { "poetry", "run", "robotcode", "language-server" }
            return
          end

          new_config.cmd = { "robotcode", "language-server" }
        end,
      },
    },
  },
}
```

This also takes care of dealing with both Poetry and uv managed repositories.


EDIT: with LazyVim 15.x, the updated configuration is:

```lua
return {
  "neovim/nvim-lspconfig",
  opts = {
    servers = {
      robotcode = {
        -- dynamically adjust cmd depending on lockfiles in root_dir
        cmd = function(dispatchers, config)
          vim.notify("robotcode root dir is: " .. config.root_dir)
          local local_cmd
          if vim.uv.fs_stat(config.root_dir .. "/uv.lock") then
            local_cmd = { "uv", "--directory", config.root_dir, "run", "robotcode", "language-server" }
          elseif vim.uv.fs_stat(config.root_dir .. "/poetry.lock") then
            local_cmd = { "poetry", "--directory", config.root_dir, "run", "robotcode", "language-server" }
          else
            local_cmd = { "robotcode", "language-server" }
          end
          -- uncomment to see what command is actually used
          -- vim.notify("Cmd used: " .. table.concat(local_cmd, " "))
          return vim.lsp.rpc.start(local_cmd, dispatchers)
        end,
      },
    },
  },
}
```

PR #9428, opened 2025-03-25
docs: add requirements for new packages

chrisgrieser:
See discussion here:
https://github.com/orgs/mason-org/discussions/2#discussioncomment-12615405

### DIFF
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -4,13 +4,16 @@
 ## Issue ticket number and link
 <!-- Leave empty if not available -->
 
+## For new packages: Evidence on requirement fulfillment (new packages only)
+<!-- A link to evidence that the requirement for the package are fulfilled. -->
+<!-- see https://github.com/mason-org/mason-registry/blob/main/CONTRIBUTING.md#requirements -->
+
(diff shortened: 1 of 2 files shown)


mehalter (approved):
A big truster in the markdown parser for enumerated lists I see 😂 this looks great! Only comment is maybe we should see if the project owner agrees or has comments?

chrisgrieser:
> A big truster in the markdown parser for enumerated lists I see 😂

Nope, I actually dislike this feature, but it's what the rest of the document uses, and I thus submit to the standard ;(

> Only comment is maybe we should see if the project owner agrees or has comments?

Yep, just wanted to open the PR already, so it's quicker for william to review.

Conarius (approved):
Let's see what William has to say

williamboman (approved):
I think this is a good start. I don't think these should be _hard_ requirements, but more of a framework for assessing acceptable packages (if the only thing a package is missing is a couple of stars it should not be a blocker).

https://docs.brew.sh/Acceptable-Formulae contains some more attributes we might want to consider, there's also an exceptions disclaimer at the bottom which I think would be healthy to adapt here as well.

PR #8383, opened 2024-12-31
feat: Add nomad and terraform

rliebz:
## Describe your changes

Added nomad and terraform, two Hashicorp tools that come with built-in `fmt` commands for files of their type. 

## Checklist before requesting a review

- [x] I have successfully tested installation of the package.
- [x] I have successfully tested the package after installation.
      

## Screenshots

image

image

image

PR #5514, opened 2024-04-28
feat(opa): add open-policy-agent cli

pbnj:
## Describe your changes

Adds [`opa`](https://www.openpolicyagent.org/docs/latest/cli/) cli for formatting, linting, building, and running rego files.

## Checklist before requesting a review

- [x] I have successfully tested installation of the package.
- [x] I have successfully tested the package after installation.

PR #9436, opened 2025-03-26
feat: Add nomad

rliebz:
## Describe your changes

Added nomad, a Hashicorp tool that comes with built-in `fmt` commands for files of their type. 

Split from: https://github.com/mason-org/mason-registry/pull/8383

## Screenshots

image

image

### DIFF
--- a/packages/nomad/package.yaml
+++ b/packages/nomad/package.yaml
@@ -1,0 +1,43 @@
+---
+name: nomad
+description: |
+  Nomad is an easy-to-use, flexible, and performant workload orchestrator that can deploy a mix of microservice,
+  batch, containerized, and non-containerized applications. Nomad is easy to operate and scale and has native Consul
+  and Vault integrations.
+homepage: https://www.nomadproject.io/
+licenses:
+  - BUSL-1.1
+languages:
+  - Nomad
+categories:
+  - Formatter
+  - Linter
+  - Runtime
+
+source:
+  # renovate:datasource=github-releases
+  id: pkg:generic/hashicorp/nomad@v1.9.7
+  download:
+    - target: darwin_x64
+      files:
+        nomad.zip: https://releases.hashicorp.com/nomad/{{ version | strip_prefix "v" }}/nomad_{{ version | strip_prefix "v" }}_darwin_amd64.zip
+      bin: nomad
+    - target: darwin_arm64
+      files:
+        nomad.zip: https://releases.hashicorp.com/nomad/{{ version | strip_prefix "v" }}/nomad_{{ version | strip_prefix "v" }}_darwin_arm64.zip
+      bin: nomad
+    - target: linux_x64
+      files:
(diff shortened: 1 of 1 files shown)


packages/nomad/package.yaml
```diff
@@ -0,0 +1,41 @@
...
+  Nomad is an easy-to-use, flexible, and performant workload orchestrator that can deploy a mix of microservice,
+  batch, containerized, and non-containerized applications. Nomad is easy to operate and scale and has native Consul
+  and Vault integrations.
+homepage: https://www.nomadproject.io/
+licenses:
+  - BUSL-1.1
+languages:
+  - Nomad
+categories:
+  - Formatter
```
mehalter, line 13:
```suggestion
  - Formatter
  - Runtime
```

I believe this is the full nomad runtime as well

mehalter (changes requested):

packages/nomad/package.yaml
```diff
@@ -0,0 +1,42 @@
...
+  - Runtime
```
mehalter, line 14:
```suggestion
  - Linter
  - Runtime
```

`nomad job validate` actually does syntax checking and can be used for linting

mehalter (changes requested):

mehalter (approved):
This looks good to me! Just need a second approval and we can get this merged in :)

Conarius:
Hm, I think adding entire tools for just a few commands might be an overkill. Also because these tools are usually installed systemwide or at least userwide. Would like to hear @williamboman's thoughts on this

mehalter:
I think this is done already. Also there is a whole category for `Runtime` for full runtimes

rliebz:
In terms of precedent, here are a couple other runtimes with similarly large scopes beyond just formatting/linting:

- Deno: https://github.com/mason-org/mason-registry/pull/61 (authored by @williamboman)
- OPA: https://github.com/mason-org/mason-registry/pull/5514 (approved by @williamboman)

For context, I often work on projects that have boilerplate nomad configuration files, which get deployed using centralized infrastructure that I don't have access to. Meaning, I'm writing and modifying nomad files while not actually operating nomad myself, and certainly not from my personal machine. Whether or not the executable has other capabilities isn't really relevant for my use case—I just need to be able to format the config files appropriately. 

For situations where I do need a tool like nomad installed locally, I use a separate version manager that can pin to specific versions on a per-directory basis that also works outside neovim, which puts those versions first in the path. Meaning, if there's a project specific version, neovim will pick that up, and if there isn't, it would fall back to the mason-installed version, which will most commonly be the latest version. The reason I opened the MR was so that my neovim could have a guaranteed formatter to fall back on if I don't have a more specific version available on that machine/directory combination.

chrisgrieser (approved):

chrisgrieser:
Makes sense, can be included imo.

PR #9508, opened 2025-03-30
feat(buf-ls): add deprecation warning to buf-ls

### LINKED ISSUE #8942: Mark buf-language-server as deprecated
Since [buf-language-server](https://github.com/bufbuild/buf-language-server) is no longer actively maintained and has been publicly archived, it should be marked as deprecated. Also, the plain buf package which is also available does mostly the same things.

Conarius:
Closes #8942

### DIFF
--- a/packages/buf-language-server/package.yaml
+++ b/packages/buf-language-server/package.yaml
@@ -10,6 +10,10 @@ languages:
 categories:
   - LSP
 
+deprecation:
+  message: This repo is no longer under active development. Please use buf instead
+  since: v0.0.0-20240213221648-8e42e5e2e24f
+
 source:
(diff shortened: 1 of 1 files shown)


Conarius:
It's also a great catch-22 situation when you need to deprecate a package and then the CI doesn't run because the package is deprecated xD

chrisgrieser:
Eh, it seems this PR cannot be merged due to the tests failing. 🙈

williamboman:
You can skip CI by using the `ci_skip` field:

```yaml
# all platforms
ci_skip: true

# only windows etc.
ci_skip:
  - win 
```

It's a bit unfortunate having lingering packages that are no longer installable. Mason doesn't handle removal of package definitions very well yet. Removing a package won't break anything, but it will cause the package to completely disappear from Mason APIs (and UIs), making it uninstallable.

PR #9597, opened 2025-04-06
ci: add stale-bot

chrisgrieser:
### Problem
We have a bunch of PRs and issues that are really old and appear to be outdated, or do not have a reaction by the PR creator after requesting changes. 

### Solution
The stale-bot could be a useful tool to reduce maintenance workload by automatically closing such issues/PRs.

I set the parameters to be very generous, an issue/PR is only considered stale after 3 months: If the creator of a PR does not implement the requested changes after such a long time, I think it is safe to assume that the creator has lost interest and abandoned the PR. (And in case that assumption is wrong, it is not a big effort for the creator to simply open a new PR.)

### DIFF
--- a/.github/workflows/stale-bot.yml
+++ b/.github/workflows/stale-bot.yml
@@ -1,0 +1,35 @@
+name: Stale bot
+on:
+  schedule:
+    - cron: "18 04 * * 3"
+
+permissions:
+  issues: write
+  pull-requests: write
+
+jobs:
+  stale:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Close stale issues/PRs
+        uses: actions/stale@v9
+        with:
+          repo-token: ${{ secrets.GITHUB_TOKEN }}
+
+          # DOCS https://github.com/actions/stale#all-options
+          days-before-stale: 90
+          days-before-close: 7
+
+          stale-issue-label: Stale
+          stale-issue-message: |
+            This issue has been automatically marked as stale.
+            If this issue is still relevant, please leave any comment, and it will be kept open.
+          close-issue-message: |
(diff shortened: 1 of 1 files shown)


Conarius (approved):
Yeah, it's very unfortunate. Though we can also edit PRs as well so we can get them merged or not, even if the contributer doesn't respond at all.

williamboman (commented):
I'm _really_ not a fan of stale bot. Almost so much that I want to introduce a veto right for @mason-org 😆. I don't see it solving any problems, just adding noise. I think we're capable of closing PRs and issues ourselves, no? Many issues and PRs are still valid even if there's been no activity in a while.

Conarius:
> I'm _really_ not a fan of stale bot. Almost so much that I want to introduce a veto right for @mason-org 😆. I don't see it solving any problems, just adding noise. I think we're capable of closing PRs and issues ourselves, no? Many issues and PRs are still valid even if there's been no activity in a while.
@williamboman Yeah, you're still the BDFL here. But I get your points. We just have to go through issues and PRs ourselves semi-regularly then, which I already do.
On another note:
I think a labeler action might be good to make it easier to look through issues.

chrisgrieser:
@williamboman I do not mind you having veto right, since it's by far still your project.

Well, I agree that the stalebot can be annoying. For instance, with folke's plugin repos, it closes issues after 30 days, which is really annoying when folke is on a vacation that is several months long. However, I do think a lot comes down to properly using the stalebot:
1. We could just use it for PRs, not issues.
2. We can configure it to be far more "generous". For instance, I initially suggested 3 months before the bot becomes active.

That way, we could use the stalebot not as a "close issues to keep the issue number down" bot, but rather as an "auto-close very old PRs that we otherwise would have to close manually" bot.

PR #9601, opened 2025-04-07
feat: add pgformatter

alensugimoto:
### Describe your changes

Adds [pgformatter](https://github.com/darold/pgFormatter), a PostgreSQL SQL syntax beautifier.

1.8k stars on GitHub.

### DIFF
--- a/packages/pgformatter/package.yaml
+++ b/packages/pgformatter/package.yaml
@@ -1,0 +1,20 @@
+---
+name: pgformatter
+description: PostgreSQL SQL syntax beautifier
+homepage: http://sqlformat.darold.net
+licenses:
+  - PostgreSQL
+languages:
+  - Postgres
+  - SQL
+categories:
+  - Formatter
+
+source:
+  # renovate:datasource=github-tags
+  id: pkg:github/darold/pgformatter@v5.6
+  build:
+    run: ""
+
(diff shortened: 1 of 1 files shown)


packages/pgformatter/package.yaml
```diff
@@ -0,0 +1,20 @@
...
+  - Postgres
+  - SQL
+categories:
+  - Formatter
+
+source:
+  # renovate:datasource=github-tags
+  id: pkg:github/darold/pgformatter@v5.6
+  build:
+    run: ""
```
Conarius, line 17: It's a bit weird that the binary is in the source code itself so it has to be installed with this but I'll allow it for now.

Conarius (approved):

mehalter:
I feel like we might want to test what architectures that binary works for. We could do a quick test by just having the build step be something like `./pgformatter --help`. Or someone could try something else

Conarius:
> I feel like we might want to test what architectures that binary works for. We could do a quick test by just having the build step be something like `./pgformatter --help`. Or someone could try something else

As far as I'm aware it uses Perl under the hood so basically any platform that Perl works on.

mehalter:
Oh true true, it's just a script. Looks good to me!

PR #9706, opened 2025-04-16
feat: add tabby-agent

dboursinos:
### Describe your changes

Adding the [tabby](https://github.com/TabbyML/tabby) LSP.

https://github.com/williamboman/mason.nvim/discussions/1360

Already exists in [nvim-lspconfig](https://github.com/neovim/nvim-lspconfig/blob/master/lua/lspconfig/configs/tabby_ml.lua).

### Screenshots

Screenshot

### DIFF
--- a/packages/tabby-ml/package.yaml
+++ b/packages/tabby-ml/package.yaml
@@ -1,0 +1,15 @@
+---
+name: tabby_ml
+description: Tabby is a self-hosted AI coding assistant, offering an open-source and on-premises alternative to GitHub Copilot.
+homepage: https://github.com/TabbyML/tabby
+licenses:
+  - Apache-2.0
+languages: []
+categories:
+  - LSP
+
+source:
+  id: pkg:npm/tabby-agent@1.8.0
+
+bin:
+  tabby-agent: npm:tabby-agent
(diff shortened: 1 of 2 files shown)


packages/tabby-ml/package.yaml
```diff
@@ -0,0 +1,15 @@
+---
+name: tabby_ml
```
Conarius, line 2: Must be the same as the path name, `tabby-ml` in this case.

Conarius (changes requested):

packages/tabby-ml/package.yaml
mehalter: Also the package name and path name should be `tabby` or `tabby-agent` based on the contribution guidelines. I believe Tabby is higher priority

dboursinos: Thank you. I changed the path and package name to `tabby-agent`.
I use `mason-lspconfig`  and verified it still works but how does it know it should use [this](https://github.com/neovim/nvim-lspconfig/blob/master/lua/lspconfig/configs/tabby_ml.lua) config from `nvim-lspconfig` when the name there is `tabby_ml`?

Conarius: That's part of mason-nvim-lspconfig which maps the LSPs with nvim-lspconfig.
Mason and mason-regsitry just installs all the tools and makes them available to Neovim.
No need to worry about the different names here.

Conarius (approved):
LGTM

chrisgrieser (approved):

Issue #9707, opened 2025-04-16
ruff-lsp is deprecated since 2025-02-06


ivuorinen:
> ruff-lsp is deprecated. Please switch to the native language server (ruff server)[...]

> ruff server supports the same feature set as ruff-lsp, but with superior performance and no installation required. ruff server was marked as stable in Ruff v0.5.3.

See:
- https://github.com/astral-sh/ruff-lsp/
- https://github.com/astral-sh/ruff-lsp/pull/520

Related files:
- https://github.com/mason-org/mason-registry/blob/main/packages/ruff-lsp/package.yaml

PR #9708, opened 2025-04-16
chore(ruff-lsp): add deprecation message

ivuorinen:
### Describe your changes

Added deprecation message to `ruff-lsp` package.yaml.

Followed example from [xo](https://github.com/mason-org/mason-registry/blob/main/packages/xo/package.yaml), but noticed at the same time [CONTRIBUTING.md](https://github.com/mason-org/mason-registry/blob/main/CONTRIBUTING.md#package-specification) package specs doesn't include the `deprecation` keyword.

- https://github.com/mason-org/mason-registry/issues/9707
- https://github.com/astral-sh/ruff-lsp/pull/520

Removed because not applicable.

### DIFF
--- a/packages/ruff-lsp/package.yaml
+++ b/packages/ruff-lsp/package.yaml
@@ -9,6 +9,10 @@ languages:
 categories:
   - LSP
 
+deprecation:
+  since: v0.0.61
+  message: `ruff server` supports the same feature set as ruff-lsp, but with superior performance and no installation required. ruff server was marked as stable in Ruff v0.5.3.
+
 source:
   id: pkg:pypi/ruff-lsp@0.0.62
 
--- a/packages/ruff-lsp/package.yaml
+++ b/packages/ruff-lsp/package.yaml
@@ -11,7 +11,7 @@ categories:
 
 deprecation:
   since: v0.0.61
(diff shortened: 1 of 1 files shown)


packages/ruff-lsp/package.yaml
```diff
@@ -9,6 +9,10 @@ languages:
 categories:
   - LSP
 
+deprecation:
+  since: v0.0.61
+  message: `ruff server` supports the same feature set as ruff-lsp, but with superior performance and no installation required. ruff server was marked as stable in Ruff v0.5.3.
```
chrisgrieser, line 14: Just to be sure, it could make sense to explicitly mention that the package is also available at mason with the name `ruff`?

ivuorinen:
```suggestion
  message: please use `ruff` as a replacement. `ruff server` supports the same feature set as `ruff-lsp`, but with superior performance and no installation required. ruff server was marked as stable in Ruff v0.5.3.
```

Could this be acceptable wording?

chrisgrieser: How about 

> please use the mason package `ruff` as replacement. The command `ruff server`

chrisgrieser (approved):

PR #9738, opened 2025-04-18
feat: use OmniSharp/OmniSharp.exe binary provided in the package

616b2f:
### Describe your changes
With the new way how Language Server Config can be provided in neovim, we changed the default binary to match the ones provided inside the package:

See official repo docs how binaries are named in the official package (`OmniSharp` and`OmniSharp.exe`):
https://github.com/OmniSharp/omnisharp-roslyn?tab=readme-ov-file#building

See new LSP config for OmniSharp:
https://github.com/neovim/nvim-lspconfig/blob/master/lsp/omnisharp.lua#L19

The benefit in doing so, would be that the LSP config does not have to change if the binary is in users PATH, and could be picked up, resulting in hopefully 'just work experience', regardless if the user installed the server manually or with Mason. 

- [x] I have successfully tested installation of the package (but only on Linux).
- [x] I have successfully tested the package after installation (but only on Linux).
- [x] I have successfully tested installation of the package (for Windows, maybe someone could test it please).
- [x] I have successfully tested the package after installation (for Windows, maybe someone could test it please).

For MacOS I expect it to work as it works for Linux.

### DIFF
--- a/packages/omnisharp/package.yaml
+++ b/packages/omnisharp/package.yaml
@@ -38,7 +38,6 @@ schemas:
 
 bin:
   OmniSharp: "{{source.asset.bin}}"
-  omnisharp: dotnet:libexec/OmniSharp.dll
 
 neovim:
   lspconfig: omnisharp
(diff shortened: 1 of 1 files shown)


616b2f:
I would keep it as Draft as long as it's not tested for Windows users.

williamboman (approved):
Tested on Windows, seems fine!

Conarius (approved):
LGTM

PR #9774, opened 2025-04-21
feat: add lspconfig mapping to packages

williamboman:
### Describe your changes
This has been something I've been wanting to do to avoid the churn in mason-lspconfig.nvim and improve the contributing
experience (no need to open a PR in mason-lspconfig.nvim, wait years for review, wait a few more years for a release,
etc.). This is accompanied with the necessary changes in mason-lspconfig.nvim (branch `2.x`), as well as the registry
schema.

I've been waiting to see how the "native" support for LSP would evolve in Neovim, and with the introduction of the
`vim.lsp.config()` API, mason-lspconfig.nvim has had its feature set reduced and tries to get out of the way as much as
possible. Going forward that plugin will essentially provide 2 main features:

- automatically enable (`vim.lsp.enable()`) installed servers
- provide convenience methods such as `:LspInstall`

More long term I'd like to deprecate mason-lspconfig.nvim altogether, and potentially move the small remaining feature
set into `mason.nvim` directly (this would probably be limited to just automatically enabling servers).

I chose to put this under the `neovim:` root key as to clearly separate it from the rest of the definition, which is
client agnostic, and prepare a "namespace" for potential future Neovim-only fields.

Related PRs & branches:
- https://github.com/mason-org/registry-schema/pull/13
- https://github.com/williamboman/mason-lspconfig.nvim/tree/2.x

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -21,6 +21,8 @@
   - [`bin`](#bin)
   - [`share`](#share)
   - [`opt`](#opt)
+  - [`neovim`](#neovim)
+    - [`lspconfig`](#lspconfig)
 - [Expressions](#expressions)
 - [Examples](#examples)
   - [Common fields](#common-fields)
@@ -278,6 +280,23 @@ opt:
     solang/llvm15.0/LICENSE: doc/LICENSE
     # Links $MASON/opt/solang/llvm15.0/ -> <package>/llvm15.0/**/* (i.e. all files within the target directory)
     solang/llvm15.0/: llvm15.0/
+```
+
+## `neovim`
+
+The `neovim` property is an object that contain metadata specific for Neovim.
+
+### `lspconfig`
+
+The `lspconfig` property of `neovim` specifies the name of the LSP configuration as provided by
+[nvim-lspconfig](https://github.com/neovim/nvim-lspconfig/). Providing the `neovim.lspconfig` value MUST be done if the
+package contains an LSP configuration entry in nvim-lspconfig.
+
+Example:
+
+```yaml
+neovim:
+    lspconfig: lua_ls
 ```
 
 # Expressions
--- a/packages/angular-language-server/package.yaml
+++ b/packages/angular-language-server/package.yaml
@@ -61,3 +61,6 @@ schemas:
 
 bin:
   ngserver: npm:ngserver
+
+neovim:
+  lspconfig: angularls
--- a/packages/ansible-language-server/package.yaml
+++ b/packages/ansible-language-server/package.yaml
@@ -14,3 +14,6 @@ source:
 
 bin:
   ansible-language-server: npm:ansible-language-server
+
+neovim:
+  lspconfig: ansiblels
--- a/packages/antlers-language-server/package.yaml
+++ b/packages/antlers-language-server/package.yaml
@@ -16,3 +16,6 @@ source:
 
 bin:
   antlersls: npm:antlersls
+
+neovim:
+  lspconfig: antlersls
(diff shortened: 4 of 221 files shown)


mehalter (approved):
This is a great addition!

mehalter:
In terms of the `mason-lspconfig` portion. Do you think it would still be useful to have the `handlers` section of the configuration to customize how language servers are setup. In most cases it could be `vim.lsp.enable`, but there are a lot of other cases where language server specific plugins do their own `vim.lsp.start` calls and setup where instead of calling `vim.lsp.enable`, you might just be calling the setup for some other plugin that does more with `vim.lsp`. It would also make it easier to say you want to automatically call `vim.lsp.enable` on all language servers apart from some exceptions

chrisgrieser:
I also like this. Having to maintain a list of servers, with mason package names as well as lspconfig names always felt cumbersome. 

A convenience method to get the lspconfig name from all installed packages in mason would be useful, so that a user can then pass that to one `vim.lsp.enable` call. (But I assume this pretty much what you mean when you say making mason-lspconfig obsolete)

mehalter:
I think all that sounds great! Yeah adding just an exclusion list for `automatic_enable` is definitely sufficient!

williamboman:
Great, I'll try to get this merged then and then cut the v2.0 release of Mason during next week (I'll give it some time for these changes in the registry to propagate).

We might have to introduce a new label to skip package tests for big PRs like these - I'll give it the CI an attempt at this PR first, but it'll most likely timeout or fail somehow.

Conarius:
Luckily nvim-lspconfig already migrated everything into the new vim.lsp.config format

williamboman:
> Luckily nvim-lspconfig already migrated everything into the new vim.lsp.config format

I don't think all of them are yet migrated. It seems like they are making progress quickly though, but I think we can adjust that in later PRs. Btw this PR is reliant on:

1. https://github.com/mason-org/actions/pull/13
1. https://github.com/mason-org/mason-registry/pull/9861

chrisgrieser:
They are almost there, looks like only 15-20 LSPs or so are still missing.
https://github.com/neovim/nvim-lspconfig/issues/3705#issuecomment-2841670795

chrisgrieser:
Nice! I just gave it a shot, and enabling LSPs now just takes a ~10 lines. 🥳 

```lua
local installedPacks = require("mason-registry").get_installed_packages()
local lspConfigNames = vim.iter(installedPacks)
	:filter(function(pack) return vim.list_contains(pack.spec.categories, "LSP") end)
	:map(function(pack)
		local lspConfigName = pack.spec.neovim and pack.spec.neovim.lspconfig ---@diagnostic disable-line: undefined-field
		if not lspConfigName then
			local msg = pack.name .. " has no `neovim` entry"
			vim.notify(msg, vim.log.levels.WARN, { title = "Mason" })
			return
		end
		return lspConfigName
	end)
	:totable()
vim.lsp.enable(lspConfigNames)
```

williamboman:
Could be further reduced to something convoluted like this 😛

```lua
local _ = require "mason-core.functional"
local registry = require "mason-registry"
local Optional = require "mason-core.optional"

local get_lspconfig_name = _.compose(Optional.of_nilable, _.path { "spec", "neovim", "lspconfig" })
local enable = _.compose(if_present(_.scheduler_wrap(vim.lsp.enable)), get_lspconfig_name)
_.each(enable, registry.get_installed_packages())
registry:on("package:install:success", enable)
```

```lua
-- pretend this is in core
_.if_present = _.curryN(function (fn, value)
    value:if_present(fn)
end, 2)
```

Gelio:
Thanks for confirming! I raised https://github.com/mason-org/mason-registry/pull/9969 to improve the protols config

PR #9775, opened 2025-04-22
fix(zk): update darwin_arm64 asset file

williamboman:
### Describe your changes

Fixes the asset file for `darwin_arm64`. I also removed `ci_skip` to see if the runners are able to unpack this. The
gzip bug seemed a bit weird, maybe it's Intel mac related (I don't have an Intel mac anymore so can't test it myself).

### DIFF
--- a/packages/zk/package.yaml
+++ b/packages/zk/package.yaml
@@ -19,7 +19,7 @@ source:
     - target: linux_x86
       file: zk-{{version}}-linux-i386.tar.gz
     - target: darwin_arm64
-      file: zk-{{version}}-macos-arm64.tar.gz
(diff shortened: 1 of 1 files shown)


williamboman:
Hm so I can actually reproduce the error. I compiled the [same version](https://github.com/apple-oss-distributions/file_cmds/tree/file_cmds-430.140.2) of `Apple gzip` that the runner is using and I'm getting

```
gzip: /Users/william/Downloads/zk-v0.15.0-macos_arm64.tar.gz: trailing garbage ignored
```

It seems to unpack the contents just fine, the issue seems to be with `gtar` which errors when it encounters a non-0 exit code from `gzip`. The more you know.

williamboman:
#9873

PR #9914, opened 2025-05-03
feat(just-lsp): add additional targets

terror:
### Describe your changes

This PR simply adds linux arm and gnu targets for the `just-lsp` lsp server. Pretty much whatever was missing from the [supported packages](https://github.com/terror/just-lsp/releases/tag/0.2.1).


chrisgrieser:
The registry lint failed because you likely entered target values that aren't valid. You can use the yaml lsp to check which ones are not valid

Conarius (approved):
LGTM

PR #9968, opened 2025-05-07
chore: add reminder to add `neovim.lspconfig` to PR template

chrisgrieser:
### Describe your changes
add reminder to add `neovim.lspconfig` to PR template. follow up to #9774


williamboman (approved):
Nice idea. Would probably be good to validate these values too somehow, but it won't cause any errors if it happens to be incorrect (for now)

PR #10126, opened 2025-05-18
feat(pint) - prometheus rule linter

WhoIsSethDaniel:
### Describe your changes
Adding the prometheus rule linter known as 'pint'. The mason package name is 'prometheus_pint' since there is already a different pint package (a php code formatter).

The tool has close to 1000 stars on Github and is maintained by Cloudflare.
It is *not* yet supported by `nvim-lint` but will be soon.

I am able to test with Linux on amd64. All other architectures are best guess.

### DIFF
--- a/packages/prometheus-pint/package.yaml
+++ b/packages/prometheus-pint/package.yaml
@@ -1,0 +1,38 @@
+---
+name: prometheus-pint
+description: Prometheus rule linter/validator
+homepage: https://cloudflare.github.io/pint/
+licenses:
+  - Apache-2.0
+languages:
+  - PromQL
+categories:
+  - Linter
+
+source:
+  id: pkg:github/cloudflare/pint@v0.73.7
+  asset:
+    - target: linux_x86
+      file: pint-{{ version | strip_prefix "v" }}-linux-386.tar.gz
+      bin: pint-linux-386
+    - target: linux_x64
+      file: pint-{{ version | strip_prefix "v" }}-linux-amd64.tar.gz
+      bin: pint-linux-amd64
+    - target: linux_arm64
+      file: pint-{{ version | strip_prefix "v" }}-linux-arm64.tar.gz
+      bin: pint-linux-arm64
+    - target: darwin_x64
+      file: pint-{{ version | strip_prefix "v" }}-darwin-amd64.tar.gz
+      bin: pint-darwin-amd64
+    - target: darwin_arm64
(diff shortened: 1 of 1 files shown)


Conarius (changes requested):

packages/prometheus_pint/package.yaml
```diff
@@ -0,0 +1,38 @@
+---
+name: prometheus_pint
+description: Prometheus rule linter/validator
+homepage: https://cloudflare.github.io/pint/
+licenses:
+  - Apache-2.0
+languages:
+  - Go
```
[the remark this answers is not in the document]
WhoIsSethDaniel: In that case it should be PromQL. I'll push the change.

Conarius (approved):
LGTM

packages/prometheus_pint/package.yaml
```diff
@@ -0,0 +1,38 @@
...
+      bin: pint-darwin-arm64
+    - target: win_x86
+      file: pint-{{ version | strip_prefix "v" }}-windows-386.tar.gz
+      bin: pint-windows-386.exe
+    - target: win_x64
+      file: pint-{{ version | strip_prefix "v" }}-windows-amd64.tar.gz
+      bin: pint-windows-amd64.exe
+
+bin:
+  prometheus-pint: "{{source.asset.bin}}"
```
chrisgrieser, line 38: Sorry for nitpicking but shouldn't it be an underscore to match the package name?

WhoIsSethDaniel: And that breaks some linting for some reason.

WhoIsSethDaniel: I don't know what to do about it.

chrisgrieser:
@WhoIsSethDaniel @Conarius (continuing the [bin name discussion here](https://github.com/mason-org/mason-registry/pull/10126#discussion_r2114336930) for visibility)

Ah true, tools like nvim-lint would now have to be passed `prometheus-lint {args}` instead of `pint {args}`, which is probably what breaks the linting.

However, the [other package has exactly `lint` as bin name](https://github.com/mason-org/mason-registry/blob/main/packages/pint/package.yaml#L16), and having both be named `pint` thus results in a conflict that will break things as soon as the user has both installed.

So whether the symlink is named `pint` or something else, it will one something either way. I am honestly what the best way to proceed here is.

WhoIsSethDaniel:
@chrisgrieser @Conarius is this waiting on me for anything?

Conarius (approved):

PR #10135, opened 2025-05-19
feat(pinact): add pinact cli

pbnj:
### Describe your changes

Add [pinact](https://github.com/suzuki-shunsuke/pinact), a utility that pins GitHub Actions & Workflows to commit SHAs instead of Git Tags to defend against software supply attacks (e.g. [tj-actions/changed-files](https://www.cisa.gov/news-events/alerts/2025/03/18/supply-chain-compromise-third-party-tj-actionschanged-files-cve-2025-30066-and-reviewdogaction))

- Project has 300+ stars on GH
- Project is actively maintained (latest commit is <24h ago)

### DIFF
--- a/packages/pinact/package.yaml
+++ b/packages/pinact/package.yaml
@@ -1,0 +1,35 @@
+---
+name: pinact
+description: Update & pin GitHub Workflows and Actions to commit SHAs
+homepage: https://github.com/suzuki-shunsuke/pinact
+licenses:
+  - MIT
+languages:
+  - YAML
+categories:
+  - Formatter
+
+source:
+  id: pkg:github/suzuki-shunsuke/pinact@v3.1.2
+  asset:
+    - target: darwin_x64
+      file: pinact_darwin_amd64.tar.gz
+      bin: actionlint
+    - target: darwin_arm64
+      file: pinact_darwin_arm64.tar.gz
+      bin: actionlint
+    - target: linux_x64
+      file: pinact_linux_amd64.tar.gz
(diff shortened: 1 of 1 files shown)


chrisgrieser:
Do I understand this correctly, this CLI changes the version to a hash? Like, in that case, I get that is somewhat works like a formatter, but using the category "formatter" feels a bit misleading to me.

Not sure whether leaving `category` empty or using `Runtime` (as kind of our residual category) is the more appropriate approach here.

pbnj:
> https://github.com/mason-org/mason-registry/blob/352d0892a1867055cdcba45cf70d675da9f32bed/packages/pinact/package.yaml#L9-L10
> 
> Do I understand this correctly, this CLI changes the version to a hash? Like, in that case, I get that is somewhat works like a formatter, but using the category "formatter" feels a bit misleading to me.
> 
> Not sure whether leaving `category` empty or using `Runtime` (as kind of residual category) is the more appropriate approach here. @mason-org/triage do we have a preference how to categorize such cases? Our [categories-spec](https://github.com/mason-org/mason-registry/blob/main/CONTRIBUTING.md#categories) doesn't go into detail here.

Happy to change this once there is confirmation on which category to use.

pbnj:
@chrisgrieser - I noticed some packages do not have a category (e.g. [gh/package.yaml](https://github.com/mason-org/mason-registry/blob/main/packages/gh/package.yaml#L10)), so I removed the category for pinact.

chrisgrieser (approved):

PR #10346, opened 2025-06-04
fix(alejandra): switch to cargo source instead of precompiled binaries

szpolny:
Precompiled binaries only target Linux (x86_64 & arm64 musl) and don’t support other platforms like aarch64-darwin. Using the cargo source ensures cross-platform compatibility.

### Describe your changes
Switched the `source` in alejandra/package.yaml from a precompiled binary to building from source using cargo (pkg:cargo/alejandra_cli@4.0.0). This change allows installation and usage on non-Linux platforms such as aarch64 (macOS)

### DIFF
--- a/packages/alejandra/package.yaml
+++ b/packages/alejandra/package.yaml
@@ -10,12 +10,6 @@ categories:
   - Formatter
 
 source:
-  id: pkg:github/kamadorueda/alejandra@4.0.0
-  asset:
-    - target: linux_x64
-      file: alejandra-x86_64-unknown-linux-musl
-    - target: linux_arm64
-      file: alejandra-aarch64-unknown-linux-musl
-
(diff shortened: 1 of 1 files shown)


szpolny:
kamadorueda (main maintainer of alejandra): 

> I can't provide pre-compiled MacOS binaries at the moment as I don't own a MacOS laptop myself nor do I own MacOS Buildkite runners

https://github.com/kamadorueda/alejandra/issues/402#issuecomment-2456046004

So this is why i created PR here.

chrisgrieser:
okay, makes sense.

chrisgrieser (approved):

mehalter:
I wonder if this would be a good use case for having some level of priority in terms of assets vs building. @williamboman is there any way for us to have the assets defined and say "if there are no assets, try to use the cargo sources"?

Issue #10189, opened 2025-05-22
Add @vue/typescript-plugin


flexagoon:
See #5064

It was closed because it was possible to use the plugin through `mason-registry`'s `get_package` method, but now that this has been removed and accessing the package directory directly is not recommended as per the Mason v2.0.0 changelog:

> -- 3. If you absolutely need to access the package directory (please consider raising an issue/PR in the registry if possible)

There should be some other way to use it (eg. through the `share` directory)

shunsuke6:
Couldn't you use the vim.fn.exepath("kotlin-debug-adapter") method described in the [changelog](https://github.com/mason-org/mason.nvim/blob/8024d64e1330b86044fed4c8494ef3dcd483a67c/CHANGELOG.md#package-api-changes)?

```diff
- require("mason-registry").get_package("vue-language-server"):get_install_path()
+ vim.fn.exepath("vue-language-server")

```

flexagoon:
@shunsuke6 that'd get the path for `vue-language-server`, which is the main executable, and it already works. However, to add a plugin to typescript, you need the full path to the installation directory, which is currently `~/.local/share/nvim/mason/packages/vue-language-server/node_modules/@vue/language-server`.

sghng:
Importing `vue` SFC inside `ts` files aren't working after upgrading to v2. I also noticed that `volar` is removed from the registry, has it become `vue_ls` now? I also found this [article](https://kosu.me/blog/breaking-changes-in-mason-2-0-how-i-updated-my-neovim-lsp-config) that addressed the issue of the typescript plugin. caused by Mason v2.

sghng:
This config solves my problem:

```lua
	{
		"mason-org/mason-lspconfig.nvim",
		dependencies = "neovim/nvim-lspconfig",
		event = "VeryLazy",
		config = function()
			require("mason-lspconfig").setup(
				---@type MasonLspconfigSettings
				{
					ensure_installed = {
						"air", -- R LSP and Formatter
						"cssls",
						"html",
						"jsonls",
						"lemminx", -- xml
						"lua_ls",
						"marksman",
						"ruff",
						"tailwindcss",
						"taplo", -- toml
						"ts_ls",
						"unocss",
						"vimls",
						"vue_ls",
					},
				}
			)

			--- Additional configuration for Vue.js support
			local vue_ls_path = vim.fn.expand("$MASON/packages/vue-language-server")
[...]

PR #10366, opened 2025-06-06
feat(vue-language-server): link package directory to share/

flexagoon:
### Describe your changes
Allows the TypeScript plugin to be used like this:

```
{
  name = "@vue/typescript-plugin",
  location = vim.fn.expand("$MASON/share/vue-language-server"),
  languages = { "vue" },
}
```

#10189

For some reason, I wasn't able to use Mason with a custom registry at all (perhaps a separate bug?). However, I manually symlinked the files and tested it like that, and everything worked.

### DIFF
--- a/packages/vue-language-server/package.yaml
+++ b/packages/vue-language-server/package.yaml
@@ -20,5 +20,8 @@ schemas:
 bin:
   vue-language-server: npm:vue-language-server
 
+share:
+  vue-language-server: node_modules/@vue/language-server
+
 neovim:
   lspconfig: vue_ls


williamboman:
Hello and thanks for the PR! This is problematic because:

1) The `vue-language-server` is technically not responsible for providing the `@vue/typescript-plugin` npm package. It is available via this path because it's a dependency of `@vue/language-server`.
2) Linking the entire `node_modules/` directory like this is problematic because of its size (2k+ files). The share directory is not really intended for this purpose.
3) It introduces/promotes an awkward (and imo incorrect) dependency between `typescript-language-server` and `vue-language-server` (the dependency is `@vue/typescript-plugin`, not transitively via `@vue/language-server`).

For now I'd recommend just specifying the path to `vue-language-server` manually, like you did before, `$MASON/packages/vue-language-server/node_modules/@vue/language-server`. Being able to use optional plugins for packages via Mason isn't really solved yet, the best/easiest solution might be to simply include the `@vue/typescript-plugin` as an extra package for `typescript-language-server` itself, although I'm not entirely sure what implications that would have. Like so:

```yaml
source:
  id: pkg:npm/typescript-language-server@4.3.4
  extra_packages:
    - typescript
    - @vue/typescript-plugin
```

PR #9776, opened 2025-04-22
Xcode build server

guzman109:
### Describe your changes
Made changes suggested here https://github.com/mason-org/mason-registry/pull/6987

PR #10585, opened 2025-06-24
feat(xcode-build-server): add xcode-build-server package

luan:
## Summary
- Add xcode-build-server package to mason-registry

## Notes
Re-creating https://github.com/mason-org/mason-registry/pull/9776 which was deleted by the original author.

### DIFF
--- a/packages/xcode-build-server/package.yaml
+++ b/packages/xcode-build-server/package.yaml
@@ -1,0 +1,22 @@
+---
+name: xcode-build-server
+description: A server to bridge Xcode build system outside of Xcode
+homepage: https://github.com/SolaWing/xcode-build-server
+licenses:
+  - MIT
+languages:
+  - Swift
+categories:
+  - LSP
+
+source:
+  id: "pkg:github/SolaWing/xcode-build-server@v1.2.0"
+  asset:
+    - target: darwin_arm64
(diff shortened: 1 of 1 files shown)


chrisgrieser:
Thanks for picking up the slack. 

Hmm, curiously now the package tests are failing which did not fail in the previous PR. I am in mobile rn, maybe the download url changed?

> Failed to download file "https://github.com/solawing/xcode-build-server/releases/download/v1.2.0/v1.2.0.tar.gz".

chrisgrieser:
Ah lol, the previous PR(s) were pointing to release artifacts that simply do not exist. https://github.com/SolaWing/xcode-build-server/releases

Looking at the readme it appears the recommended way of installing the tool is pretty much "clone repo and create symlink", a scenario that I think has not been considered yet by the mason spec? 

I think something like fetching the zipped/targz source code from the release artifacts might work.

Conarius:
@chrisgrieser I think the only solution would be to just state all supported platforms and then do the bin symlink iirc.

luan:
Humm I see; I hadn't even realised that was a problem, perhaps that's why the original author gave up on this, do y'all have any idea what we could do here?

Conarius:
Already merged it in a newer PR, still thanks!

PR #10669, opened 2025-07-02
feat: add emmylua-codeformat

yutkat:
### Describe your changes

I would like to add the EmmyLuaCodeStyle formatter, which is also included in lua_ls.

https://github.com/CppCXY/EmmyLuaCodeStyle

swappy-20250702-102015


chrisgrieser:
> which is also included in lua_ls

Is there a reason to install this formatter when you already get it via the lua-ls?

yutkat:
In my personal view, there are cases where lua-ls takes a long time to start up, so I would like to use only the formatter.

Conarius (approved):
LGTM

Issue #10028, opened 2025-05-11
wgsl-analyzer is out of date


RegenJacob:
It seems like wgsl-analyzer is not up to date.
This could be because the release naming scheme of wgsl-analyzer [changed](https://github.com/wgsl-analyzer/wgsl-analyzer/releases) making the renovate bot not capable of updating it.

This might be related to #9832

Conarius:
The update and renovate string were updated, should work now for future updates.

PR #10979, opened 2025-07-27
bump(wgsl-analyzer): update to v2025

ov-kevin:
### Describe your changes

Bumps the `wgsl-analyzer` version the same way that renovatebot does in https://github.com/mason-org/mason-registry/pull/10977, but also updates the `assets` mappings to the types in the newest releases.

The `bin` mapping is also changed from `wgsl_analyzer` to `wgsl-analyzer` to reflect the config for `wgsl-analyzer` in lspconfig (see https://github.com/neovim/nvim-lspconfig/blob/master/lsp/wgsl_analyzer.lua). This may cause a breaking change in some lsp configs using mason, but should also enable `wgsl-analyzer` to work without any changes to the lsp config `cmd` value on a new config.

#10028 

### DIFF
--- a/packages/wgsl-analyzer/package.yaml
+++ b/packages/wgsl-analyzer/package.yaml
@@ -36,6 +36,20 @@ source:
       file: wgsl-analyzer-aarch64-pc-windows-msvc.zip
       bin: wgsl-analyzer.exe
 
+  version_overrides:
+    - constraint: semver:<=v0.9.11
+      id: pkg:github/wgsl-analyzer/wgsl-analyzer@v0.9.11
+      asset:
+        - target: [darwin_arm64, darwin_x64]
+          file: wgsl-analyzer-darwin-x64
+          bin: wgsl-analyzer-darwin-x64
+        - target: linux_x64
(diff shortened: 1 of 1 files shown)


Conarius (commented):
Maybe version constraints for 0.9.11 would be great so when someone needs the older version, they can still get them
Other than that, thank you a lot!

ov-kevin:
@Conarius Thanks for the feedback! I have added version constraints for versions <=v0.9.11. I have tested this by installing both `wgsl-analyzer@v0.9.11` and `wgsl-analyzer@2025-06-28`. In both cases the binaries are being installed in the right directory and both versions exhibit the expected behaviour when attached to a `.wgsl` buffer.

ov-kevin:
Additionally (though not directly relevant for the review of this PR), as for renaming the `wgsl_analyzer` executable to `wgsl-analyzer`, I have found the following issue https://github.com/wgsl-analyzer/wgsl-analyzer/issues/394, where using `wgsl_analyzer` is being seen as a bug, so I think renaming the executable should be alright.

Conarius (approved):
LGTM

PR #11047, opened 2025-08-01
bump(fennel-ls): from 0.1.3-4 to 0.2.1-1

gbroques:
See the `fennel-ls` package on Luarocks for available versions:
https://luarocks.org/modules/xerool/fennel-ls

PR #11057, opened 2025-08-02
bump(fennel-ls): update to 0.2.1-2 and remove ci skip

XeroOl:
### Describe your changes
The `fennel-ls` rockspec on luarocks now declares that it doesn't support windows,
so that information doesn't need to be duplicated in mason-repository.

https://github.com/mason-org/mason-registry/pull/11047

https://luarocks.org/manifests/xerool/fennel-ls-0.2.1-2.rockspec

### DIFF
--- a/packages/fennel-ls/package.yaml
+++ b/packages/fennel-ls/package.yaml
@@ -10,20 +10,10 @@ categories:
   - LSP
 
 source:
-  id: pkg:luarocks/fennel-ls@0.2.1-1
+  id: pkg:luarocks/fennel-ls@0.2.1-2
 
 bin:
   fennel-ls: luarocks:fennel-ls
 
 neovim:
   lspconfig: fennel_ls
-
-ci_skip:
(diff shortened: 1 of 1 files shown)


Conarius:
Huh, weird. It correctly says it doesn't support windows. Maybe `supported_platforms` should suffice in this case.

chrisgrieser:
Maybe changes at luarocks need a bit before they are applied due to caching?

chrisgrieser (approved):

williamboman:
Ah, I completely oversaw the CI behavior when a platform is not supported, for some reason I assumed this specific error would be handled, which it's not (Mason just passes the Luarocks error forward).

Let's just keep the `ci_skip` for now, other options would require extending Luarocks support in mason.nvim to handle such cases.

edit: The `supported_platforms` approach would be possible with https://github.com/mason-org/mason.nvim/pull/2002.

Issue #11059, opened 2025-08-02
Add docker-language-server


artem-nefedov:
Add support for installing an official [docker language server](https://github.com/docker/docker-language-server).

Differences compared to previously existing [dockerfile language server](https://github.com/rcjsuen/dockerfile-language-server):

1. Written in go, not typescript.
2. Supports Dockerfile, docker-compose.yaml, and docker-bake.hcl.
3. Owned by Docker org.

[Already supported in nvim-lspconfig.](https://github.com/neovim/nvim-lspconfig/commit/e3d837b938bd1d7e84c3eb0362b6761e4d5dc15d)

Conarius:
Has been merged with lspconfig link

PR #11061, opened 2025-08-02
feat(docker-language-server): add lsp server

chrisands:
### Describe your changes

Added docker-language-server

#11059

- [ ] At least 100 stars on GitHub.
  - at the moment 81 https://github.com/docker/docker-language-server
- [ ] Evidence that the tool has a relevant user base, such as VSCode marketplace downloads.
  - Not sure here. Package is published by docker org
- [x] Be approved at [nvim-lspconfig](https://github.com/neovim/nvim-lspconfig).
  - Already merged into master https://github.com/neovim/nvim-lspconfig/commit/e3d837b
- [x] Have a credible authority vouch for it, e.g., the tool is officially recommended by maintainers of the language.
  - Package is published by docker org on 16 july because of that usage is inevitable 

  Successfully started server inside nvim and got diagnostics for Dockerfile from server

### DIFF
--- a/packages/docker-language-server/package.yaml
+++ b/packages/docker-language-server/package.yaml
@@ -1,0 +1,20 @@
+---
+name: docker-language-server
+description: Language server for Dockerfiles, Compose files, and Bake files.
+homepage: https://github.com/docker/docker-language-server
+licenses:
+  - MIT
+languages:
+  - Docker
+categories:
+  - LSP
+
+source:
+  id: pkg:golang/github.com/docker/docker-language-server/cmd/docker-language-server@v0.14.0
+
+bin:
+  docker-language-server: golang:docker-language-server
+
+neovim:
+  lspconfig: docker-language-server
+
--- a/packages/docker-language-server/package.yaml
+++ b/packages/docker-language-server/package.yaml
@@ -3,17 +3,30 @@ name: docker-language-server
 description: Language server for Dockerfiles, Compose files, and Bake files.
 homepage: https://github.com/docker/docker-language-server
 licenses:
-  - MIT
+  - Apache-2.0
 languages:
   - Docker
 categories:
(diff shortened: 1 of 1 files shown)


chrisgrieser (commented):
Does this LSP require `go` to work, or can you also use it without `go` using the release binaries? 

In case of the latter, I'd suggest installing via fetching the release artifacts rather than via golang, since that way we can avoid having the go toolchain as a dependency for this lsp.

ebkalderon:
Not the PR author, but it appears that this language server does not require the Go toolchain to be installed if you are running from the published release binaries. Go is only required if you choose to build the language server from source, if there is no precompiled binary available for your architecture on GitHub Releases.

AFAICT, the release binaries for `docker-language-server` seem to be the same as those used in the [Docker DX VSCode extension](https://github.com/docker/vscode-extension), which itself doesn't require Go to be installed on the user's machine in order to work. [See this script to learn how it downloads the binaries from GitHub Releases](https://github.com/docker/vscode-extension/blob/43dac4f/build/downloader.mjs#L69-L86).

chrisands:
@chrisgrieser thanks for noticing. Updated configuration to use github release artifacts.

chrisgrieser (commented):
The tests say that the yaml is invalid. Check out the file with the yaml lsp installed and you should see what needs to be fixed, I think some of the strings need to be quoted.

Conarius:
Yeah, the `{{source.asset.file}}` needs to have quotes around it aka it does need to be a string.

Conarius (approved):
LGTM

ebkalderon:
Thanks so much for submitting this, @chrisands! It was perfect timing, seeing this proposal show up in my feed when I was going to reconfigure my Neovim setup for a new Docker Compose project, haha.

PR #11171, opened 2025-08-10
bump(ocaml-lsp): update to v1.23.0

czrd:
update ocaml-lsp to v1.23.0 because v1.22.0 has some issues like https://github.com/ocaml/ocaml-lsp/issues/1531

https://github.com/ocaml/ocaml-lsp/releases/tag/1.23.0


Conarius (approved):
LGTM

PR #8155, opened 2024-12-07
feat: add vsg formatter

Conarius:
## Describe your changes
Add vsg formatter for VHDL files

## Checklist before requesting a review

- [x] I have successfully tested installation of the package.
- [x] I have successfully tested the package after installation.

PR #11180, opened 2025-08-11
fix!: remove duplicate package for `vsg`

chrisgrieser:
As @fredrikfoss [pointed out](https://github.com/mason-org/mason-registry/pull/8155#issuecomment-3174710564), PR #8155 added a duplicate package, since `vsg` is already available at mason but under a different name:
- [vsg](https://github.com/mason-org/mason-registry/blob/main/packages/vsg/package.yaml)
- [vhdl-style-guide](https://github.com/mason-org/mason-registry/blob/main/packages/vhdl-style-guide/package.yaml)

I removed the latter one, since we generally use the binary names as package names. (Which is kind of a useful rule I guess, since it avoids duplications such as these...)


Conarius (approved):
LGTM

williamboman:
I've refrained from removing packages from the registry, mainly because of how the mason.nvim client handles it (it doesn't). People who have this package installed will no longer be able to manage it - i.e. it'll remain installed on the filesystem with no interface to uninstall it other than cleaning up files manually (no bueno). It might also cause some issues with the UI where it detects the package to be installed on the fs, but when it tries to access the package metadata it's not available in the registry and errors.

This is really a limitation with mason.nvim itself, not the registry, and mostly due to implementation details - I'm all for removing packages when justified. On 📱 atm so can't verify the exact behavior, but I'd prefer to pause this until mason.nvim is patched to handle it better

chrisgrieser:
Bump

PR #11327, opened 2025-08-23
feat(google-java-format): use binary executable when possible

bydrim:
### Describe your changes

previously only jar was used even though google-java-format repository already provides some binary executables for some operating systems. jar is too slow (to start) compared to binary executables. for example, if you autoformat on save, it usually times out. so it makes sense to use binary executable whenever possible.

### DIFF
--- a/packages/apex-language-server/package.yaml
+++ b/packages/apex-language-server/package.yaml
@@ -12,7 +12,7 @@ categories:
   - LSP
 
 source:
-  id: pkg:github/forcedotcom/salesforcedx-vscode@v64.10.1
+  id: pkg:github/forcedotcom/salesforcedx-vscode@v64.11.1
   asset:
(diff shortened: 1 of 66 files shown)


packages/google-java-format/package.yaml
```diff
@@ -12,7 +12,49 @@ categories:
...
+    - target: win_x64
+      file: google-java-format_windows-x86-64.exe
+      bin: google-java-format_windows-x86-64.exe
+    # All other platforms that don't have binary
+    - target: [
+        unix,
+        darwin,
+        darwin_x64,
+        linux,
+        linux_arm,
```
chrisgrieser, line 33: Not totally sure about it, but I think you don't need to list every platform here, I think the generic ones like `linux` suffice. 

(You can test it by keeping it your own platform and see what artifact gets downloaded for you)

williamboman (approved):
LGTM

PR #11533, opened 2025-09-12
Add package rpm-spec-language-server

guillermodotn:
### Describe your changes
New package with LSP for RPMs spec-files.

The LSP is present at `nvim-lspconfig`: https://github.com/neovim/nvim-lspconfig/blob/master/lsp/rpmspec.lua

Successful start of LSP in Neovim.

### DIFF
--- a/packages/rpm-spec-language-server/package.yaml
+++ b/packages/rpm-spec-language-server/package.yaml
@@ -1,0 +1,19 @@
+---
+name: rpm-spec-language-server
+description: Language server protocol (LSP) support for RPM Spec files.
+homepage: https://github.com/dcermak/rpm-spec-language-server
+licenses:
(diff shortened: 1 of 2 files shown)


packages/rpm-spec-language-server/package.yaml
```diff
@@ -0,0 +1,19 @@
+---
+name: rpm-spec-language-server
```
chrisgrieser, line 2: The name of the package should in general be the same as the one of the binary, that is `rpm_lsp_server`. (Note that the name of the parent folder needs to match that as well.)

chrisgrieser (changes requested):

chrisgrieser (approved):

Conarius (approved):
LGTM

PR #11983, opened 2025-10-20
feat(laravel-ls): add darwin binaries

pnx:
### Describe your changes
update laravel-ls to 0.0.11 and also include darwin binary.

### DIFF
--- a/packages/laravel-ls/package.yaml
+++ b/packages/laravel-ls/package.yaml
@@ -16,6 +16,10 @@ source:
       file: laravel-ls-{{version}}-linux-amd64
     - target: win_x64
       file: laravel-ls-{{version}}-windows-amd64
+    - target: darwin_x64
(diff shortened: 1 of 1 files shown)


Conarius (changes requested):
You don't need to change the version number, it's already updated as it's done automatically.
But the darwin binary is a good addition!

pnx:
> You don't need to change the version number, it's already updated as it's done automatically. But the darwin binary is a good addition!

Oh right. i see the automatic PR now :) had problems before when the releases was marked as prereleases. Updated PR. this PR should be merged after v0.0.12

Conarius:
Just merge the current main into the branch and the CI/CD should be green everywhere.

pnx:
Updated. CI/CD now passes.

Conarius (approved):
LGTM

PR #12094, opened 2025-10-29
feat: add postgres-language-server

psteinroe:
### Describe your changes

Adding `postgres-language-server`.

The project itself is already in the registry as `postgrestools`, but we are phasing out the old name due to naming and license issues. We are dual-publishing for a while, so no hurry.

I could not find any infos on how to handle such cases. Adding the new name as a new package and then deleting the old one once its phased out seemed to be the easiest way forward, but happy to do it differently.

https://github.com/supabase-community/postgres-language-server


Conarius:
I'd say the plan moving forward would be that as soon as `postgres-tools` is phased out, it should be marked as deprecated

PR #12099, opened 2025-10-29
chore: deprecate postgrestools

psteinroe:
### Describe your changes
this is the follow up for #12094


chrisgrieser (approved):
yeah, we do not have a procedure for packages being renamed, so a new package under the different name, and deprecating the old one seems to be the most reasonable approach at the moment.

Conarius (approved):
LGTM

PR #8209, opened 2024-12-12
feat: add pkl-lsp

mehalter:
## Describe your changes
This adds the relatively new official [pkl-lsp](https://github.com/apple/pkl-lsp) directly from Apple.

## Checklist before requesting a review

- [x] I have successfully tested installation of the package.
- [x] I have successfully tested the package after installation.

PR #12135, opened 2025-10-31
docs: improve phrasing for inclusion requirements

chrisgrieser:
### Problem
Our inclusion criteria are meant to be only fulfilled by packages that have evidence of active usage. However, with the current phrasing of the last criterion, a tool that is created by the solo-creator of a language, but where the language itself has little usage, technically fulfills the criterion.

See #11962 

### Suggested Solution
Change the phrasing from "recommended by credible authority like the language maintainer" to "recommended by reputable consortium, foundation, or company". That way packages for niche languages do not fulfill the requirements, but packages like for example `pkl_lsp` by Apple, which at the time of the PR (#8209) did not fulfill the other requirements yet, pass the inclusion criterion.

I think this phrasing better matches the *intention* of the requirements to not include niche tools for security reasons, since tools by companies like Apple or foundations like the Mozilla Foundation are much less unlikely to pose a security problem.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -52,6 +52,9 @@ At least one of the following conditions MUST be fulfilled for new packages to b
 1. Be approved at [nvim-lspconfig](https://github.com/neovim/nvim-lspconfig).
 1. The tool is officially recommended by a reputable consortium, foundation, or company.
 
+Even if one, or all, of the previous conditions are fulfilled, new package additions may be rejected for other 
+reasons.
+
 # Introduction
 
 * Make sure to follow the [naming guidelines](#name).


Conarius (approved):
LGTM

williamboman (approved):
LGTM. I don't think these should be viewed as a guarantee for approval, there can always be exceptions (I like how Homebrew describes this: https://docs.brew.sh/Acceptable-Formulae#sometimes-there-are-exceptions). Might be worth including something similar to set expectations right. To keep it short:

> Even if one, or all, of the following conditions are fulfilled, new package additions may be rejected for other reasons.

chrisgrieser:
@williamboman good idea, added the sentence as extra clause.

PR #12226, opened 2025-11-07
feat(pgformatter): 5.8

jay-babu:
### Describe your changes

Updating pgformatter

https://github.com/darold/pgFormatter

### DIFF
--- a/packages/pgformatter/package.yaml
+++ b/packages/pgformatter/package.yaml
@@ -12,7 +12,7 @@ categories:
 
 source:
   # renovate:datasource=github-tags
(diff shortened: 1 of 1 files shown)


jay-babu:
@chrisgrieser There has to be a bug then. There’s been two releases since 5.6 which were months ago. https://github.com/darold/pgFormatter/tags

Conarius:
Ah, it uses the github tags instead of the releases, which might be the issue here

jay-babu:
> Ah, it uses the github tags instead of the releases, which might be the issue here

Is that not supported or something?

PR #12270, opened 2025-11-11
fix(rumdl): add top level `bin` entry for rumdl executable

rodriguezjasonlloyd:
Ensure rumdl is symlinked into mason/bin by adding top level `bin` entry

- [/] I have successfully tested installation of the package.
- [/] I have successfully tested the package after installation.


Conarius (approved):
LGTM

PR #12567, opened 2025-12-02
Rollback Lua language server version to 3.15.0

pyoif:
### Describe your changes
Rollback Lua Language Server version to 3.15.0 prevent issue in 3.16.0

Fixes LuaLS/lua-language-server#3301

### DIFF
--- a/renovate.json5
+++ b/renovate.json5
@@ -24,6 +24,7 @@
     ],
     "ignoreDeps": [
         "rescript-lsp",
-        "cuepls"
+        "cuepls",
+        "lua-language-server"
     ]
 }


pyoif:
> Hmm, looking at the issue, it seems only Linux users report an issue. Also, the report is pretty recent, so it might be that the lua_ls maintainers push a fix soon.
> 
> Feels a bit hasty to me to downgrade everyone for an issue that only affects some and might potentially be fixed within a few hours?
> 
> You can run `:MasonInstall lua-language-server@3.15.0` to downgrade in the meantime, and if you use a plugin to auto-update mason packages, those usually allow pinning to a specific version.

Umm... I forget that mason can install specific version 😅 thanks for mention it, also i hope sumneko or someone fix it quickly, i'll close this PR, thanks

badosu:
> Hmm, looking at the issue, it seems only Linux users report an issue. Also, the report is pretty recent, so it might be that the lua_ls maintainers push a fix soon.
> 
> Feels a bit hasty to me to downgrade everyone for an issue that only affects some and might potentially be fixed within a few hours?
> 
> You can run `:MasonInstall lua-language-server@3.15.0` to downgrade in the meantime, and if you use a plugin to auto-update mason packages, those usually allow pinning to a specific version.

Had the same problem, I can confirm the workaround works.

PR #12590, opened 2025-12-06
feat: add wc-language-server

break-stuff:

### DIFF
--- a/packages/wc-language-server/package.yaml
+++ b/packages/wc-language-server/package.yaml
@@ -15,7 +15,7 @@ categories:
   - LSP
 
 source:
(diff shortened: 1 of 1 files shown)


chrisgrieser:
Inclusion requirements are only fulfilled if the lsp is admitted to nvim-lspconfig https://github.com/neovim/nvim-lspconfig/pull/4228

break-stuff:
@chrisgrieser it has been merged

chrisgrieser:
Yes, only needs to pass the tests now

Conarius:
It seems to me that the quotation marks might be the problem as it automatically puts it around it when parsing so it can't escape strings anymore.
At leats that's my theory.

PR #12591, opened 2025-12-07
fix(postgrestools): removed invalid lspconfig

BasicPlains:
### Describe your changes
Updates to the `postgres_lsp` configuration in `neovim/nvim-lspconfig` now point to `postgres-language-server`, and `vim.lsp.enable('posgres_lsp')` no longer works with the `postgrestools` package.

https://github.com/neovim/nvim-lspconfig/pull/4155

Additionally, since both the `postgrestools` and `postgres-language-server` contain the same lspconfig, there is an issue with `mason-lspconfig`'s `ensure_installed`.

I don't know the easiest way to show this but the gist of it is `ensure_installed` gets the list of packages from `mason-lspconfig.mappings.get_mason_map()`, which uses `mason-core.functional.invert` to get the lspconfig -> package conversions. Since the `invert` function doesn't expect the duplication, `ensure_installed` installs whichever package happened to be farther down in the local registry. Within a few restarts of neovim you'll have both `postgrestools` and `postgres-language-server` installed. If between editing your config and opening a file `mason-lspconfig` only installed the deprecated `postgrestools` you won't have any active LSPs.

I successfully tested the updated package.yaml with
```
vim.lsp.config('postgres_lsp', { cmd = { "postgrestools", "lsp-proxy" } } )
```

### DIFF
--- a/packages/postgrestools/package.yaml
+++ b/packages/postgrestools/package.yaml
@@ -32,6 +32,3 @@ source:
 
 bin:
   postgrestools: "{{source.asset.file}}"
-
-neovim:
-  lspconfig: postgres_lsp


chrisgrieser (approved):
The removal of the lsp config is straightforward, but the issue of duplicate packages is a bit trickier. Right now, the registry has no way of removing packages properly. Best you can do is mark one of them as deprecated.

however, if only one package contains a `neovim.lspconfig`, mason-lspconfig.nvim should not display that behavior, I think.

Conarius (approved):
LGTM

PR #12633, opened 2025-12-07
fix(ast-grep): switch to PyPI version of the package

Jorenar:
### Describe your changes

Due to recent Shai-Hulud 2.0 attack on NPM repositories, many places recommend or mandate disabling install scripts. The NPM version of `ast-grep` package uses the postinstall step as a workaround for selecting correct libraries for the system, which does not happen with scripts disabled, resulting in a dummy being left instead. The solution is to switch to version from PyPI, which has no such problems.

### DIFF
--- a/packages/ast-grep/package.yaml
+++ b/packages/ast-grep/package.yaml
@@ -28,11 +28,11 @@ categories:
   - LSP
 
 source:
-  id: pkg:npm/%40ast-grep/cli@0.40.0
+  id: pkg:pypi/ast-grep-cli@0.40.0
 
 bin:
-  ast-grep: npm:ast-grep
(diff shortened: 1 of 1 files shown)


chrisgrieser:
I am slightly hesistant about switching from npm to pypi, because for users who do not have (a sufficiently recent) python on their system, this PR will break their installation of ast-grep without them knowing why.

Alternatively, the package could also directly download the github releases, that way we would not be dependent on npm or pypi, and it would work for everyone?

Jorenar:
Yeah, grabbing it directly from GH would be even better!

Jorenar:
PR #12634 is with downloading as GitHub asset

PR #12634, opened 2025-12-07
fix(ast-grep): download directly as GitHub asset

Jorenar:
### Describe your changes

Due to recent Shai-Hulud 2.0 attack on NPM repositories, many places recommend or mandate disabling install scripts. The NPM version of `ast-grep` package uses the postinstall step as a workaround for selecting correct libraries for the system, which does not happen with scripts disabled, resulting in a dummy being left instead. Downloading it directly from GitHub avoids the issue.

### DIFF
--- a/packages/ast-grep/package.yaml
+++ b/packages/ast-grep/package.yaml
@@ -28,11 +28,29 @@ categories:
   - LSP
 
 source:
-  id: pkg:npm/%40ast-grep/cli@0.40.0
+  id: pkg:github/ast-grep/ast-grep@0.40.0
+  asset:
+    - target: darwin_arm64
+      file: app-aarch64-apple-darwin.zip
+    - target: darwin_x64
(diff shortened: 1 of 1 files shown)


chrisgrieser (approved):
Looks good. 

The failing Windows tests are probably due to lack of handling of the `.exe` ending.

Jorenar:
> The failing Windows tests are probably due to lack of handling of the `.exe` ending.

Indeed. Would something like this work?
```
    - target: linux_x64_gnu
      file: app-x86_64-unknown-linux-gnu.zip
    - target: win_x86
      file: app-i686-pc-windows-msvc.zip
      ext: .exe

bin:
  ast-grep: ast-grep{{source.asset.ext}}
  sg: sg{{source.asset.ext}}
```

chrisgrieser:
Yep, works. LGTM, now needs only a second review

PR #12914, opened 2025-12-26
feat: added systemd-lsp

magnuslarsen:
### Describe your changes

Added the [systemd-lsp](https://github.com/JFryy/systemd-lsp) sprung-off of the not-so-maintained [systemd-language-server](https://github.com/neovim/nvim-lspconfig/issues/4134)

https://github.com/neovim/nvim-lspconfig/blob/master/doc/configs.md#systemd_lsp

### DIFF
--- a/packages/systemd-lsp/package.yaml
+++ b/packages/systemd-lsp/package.yaml
@@ -1,0 +1,29 @@
+---
+name: systemd-lsp
+description: a language server implementation for systemd unit files made in rust
+homepage: https://github.com/JFryy/systemd-lsp
+licenses:
+  - MIT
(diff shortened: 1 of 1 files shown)


magnuslarsen:
> > the not-so-maintained systemd-language-server
> 
> Then I think the old systems lsp should be marked as deprecated s as well in this PR

I can definitely do that tomorrow 

But "the old one" is not officially unmaintained, just not updated. nvim-lspconfig still have both up and running, and the thread linked above about deprecating / removing the old one has not reached any conclusion yet... So I don't want to be too trigger happy on deprecating it :-)

PR #12963, opened 2025-12-30
feat: add Xcode-build-server

guzman109:
### Describe your changes
Adding Xcode-build-server to aide in iOS development

https://github.com/SolaWing/xcode-build-server
Over 100 stars

### DIFF
--- a/packages/xcode-build-server/package.yaml
+++ b/packages/xcode-build-server/package.yaml
@@ -8,10 +8,11 @@ languages:
   - Swift
 categories:
   - LSP
-
 source:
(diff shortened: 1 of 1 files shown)


Conarius:
I think you can just add the xcode-build-server binary in the bin field as it will be automatically linked. No need to do that for the build

guzman109:
> I think you can just add the xcode-build-server binary in the bin field as it will be automatically linked. No need to do that for the build

I actually tried that, but I could not get it to work. I'll attempt it again.

Conarius (approved):
LGTM

PR #13014, opened 2026-01-04
fix: add correct actions-language-server bin entry

williamboman:


packages/gh-actions-language-server/package.yaml
```diff
@@ -13,7 +13,10 @@ source:
   id: pkg:npm/%40actions/languageserver@0.3.32
 
 bin:
+  # gh-actions-language-server is the old executable name from the previous 3rd party distribution.
+  # It's retained for now to avoid breaking configurations.
   gh-actions-language-server: npm:actions-languageserver
+  actions-language-server: npm:actions-languageserver
```
616b2f: Would it make sense to keep the name of the original binary? That way the lsp config could be agnostic regardless of the installation way (local manual install or via mason)

PR #13039, opened 2026-01-05
fix: add supported_platforms back to nixfmt

deirn:
### Describe your changes

Hey all.
I'm the author of [mason.el](https://github.com/deirn/mason.el), which use the registry to install packages for Emacs.

I have a CI that runs daily for validating my project against all the packages from the registry, which failed to resolve the recipe for nixfmt because it can't get the target file for MacOS.

The CI currently only ignores the error if the recipe's `supported_platform` doesn't contain the current platform:
https://github.com/deirn/mason.el/commit/64f6842da0b4505e2ebfa0e2d980c1f1b1f969ea
https://github.com/deirn/mason.el/actions/runs/20724712516/job/59497294737

Changing it would make the logic more complex as I would've liked, so I think changing it here will be better.

Cheers.

### DIFF
--- a/packages/nixfmt/package.yaml
+++ b/packages/nixfmt/package.yaml
@@ -14,6 +14,8 @@ source:
   asset:
     - target: linux_x64
       file: nixfmt-x86_64-linux
+  supported_platforms:
+    - linux_x64
 
 bin:
   nixfmt: "{{source.asset.file}}"


deirn:
Thanks!

Found another way to detect the platform error by matching the log, I forgot that it already saves it for each package for the package info screen :sweat_smile:: https://github.com/deirn/mason.el/commit/03d950a67301f0436cbf16f22da434e2b4ab14a3 

> wouldn't all other packages that don't define an entry for macOS fail the CI too

I guess so, I only just added the MacOS CI, and it only for arm64 and I guess it only nixfmt that doesn't support it.
I ran it locally by spoofing my platform and it found more packages for x64.

Closing this.

PR #13068, opened 2026-01-07
chore: update deprecation entries to use date format

williamboman:


williamboman:
@mason-org/triage WDYT? Relevant PR for the schema https://github.com/mason-org/registry-schema/pull/22

chrisgrieser (approved):
Technically, those were deprecated at some point in the past, but for all practical intents and purposes, using today's date is completely fine, so all good for me.

williamboman:
> Technically, those were deprecated at some point in the past, but for all practical intents and purposes, using today's date is completely fine, so all good for me.

Yeah I figured it'd be easier to track when it's "ok" to potentially remove these by resetting their deprecation date to today.

Added some docs too. Not sure why the lint worfklow doesn't consider `2026-01-07` to be a string, will look into that next.

PR #13163, opened 2026-01-14
(bzl) Add deprecation notice

samrjack:
### Describe your changes
The bzl server has been broken for a while now. I don't know exactly when it broke but I can't download it. The home page https://bzl.io is also down. It makes sense to deprecate the server so others don't try to install it.

Since it's broken and unusable, I think it's a good idea to remove it once the 3 months deprecation requested by the CONTRIBUTING.md file are up.

Screenshot 2026-01-13 at 10 45 24 PM

### DIFF
--- a/packages/bzl/package.yaml
+++ b/packages/bzl/package.yaml
@@ -41,3 +41,6 @@ bin:
 
 neovim:
   lspconfig: bzl
+
+# Package is no longer installable.
+ci_skip: true


Conarius (approved):
LGTM

samrjack:
Thanks for taking a look! I think this one will require an override of some sort since the LSP source is down so all package tests are failing.

Conarius:
Usually a CI-skip: true should be enough

samrjack:
Oh, like [here](https://github.com/mason-org/mason-registry/pull/9508/files)! I'll update it.

PR #13164, opened 2026-01-14
[WIP] Support for clangd on ARM64

alexghergh:
### Describe your changes

This is a tentative fix for clangd support on ARM64.

[Clangd](https://github.com/clangd/clangd/releases) itself doesn't publish any ARM64 package, and so Mason fails with "Unsupported platform". 

A workaround is to use the [LLVM](https://github.com/llvm/llvm-project/releases/) release and obtain clangd from there under ARM64 (this PR shows the package format, tested and working), however I am not certain how it would be possible to combine both sources, other than having different packages (e.g. "clangd" and "clangd-arm" or something like that in the mason registry).

It might be also a possibility to use the LLVM releases for all other platforms, though it appears to me that macOS doesn't have x86 support there (only Apple Silicon).

Is there any way to combine multiple sources under one single package in the mason registry, that I missed?


Conarius:
Wait, please don't delete any other packages willy nilly, especially every other package than clangd.

alexghergh:
This is meant to just be a proof of concept (which doubles as personal-use). The PR is not mergeable anyway until the questions above can be answered.

Should've made a different branch for personal use, I apologize.

Myriade-Medieval:
@alexghergh you can set this PR to be a draft, and also if you don't want to have the other yaml files locally (deleting them from your repo doesn't delete them from the .git folder so I don't think it's useful), you could do a sparse checkout

williamboman:
Closing due to large diff (unreviewable) and WIP tag, feel free to reopen when it's ready.

PR #13253, opened 2026-01-20
feat(snyk): consolidate packages and update supported languages

nick-y-snyk:
### Describe your changes
This PR consolidates the `snyk` and `snyk-ls` packages into a single `snyk` package that serves as both an LSP and Linter (similar to `shopify-cli`).

Changes:
- Merged `snyk-ls` package functionality into `snyk` package
- Added LSP category alongside Linter
- Updated languages list to include all Snyk-supported languages based on [official documentation](https://snyk.io/product/snyk-code/):
  - Added: .NET, Apex, C#, C++, Dart, Elixir, Groovy, Java, Kotlin, PHP, Scala, Swift
  - Kept existing: Docker, Go, Helm, JavaScript, Python, Ruby, Rust, Terraform, TypeScript
- Added neovim lspconfig reference for `snyk_ls`
- Removed separate `snyk-ls` package to avoid duplication

- [x] Language names follow mason-registry naming conventions

### DIFF
--- a/packages/snyk-ls/package.yaml
+++ b/packages/snyk-ls/package.yaml
@@ -1,0 +1,39 @@
+---
+name: snyk-ls
+description: |
+  The language server follows the Language Server Protocol and integrates with Snyk Open Source, Snyk Infrastructure as
+  Code and Snyk Code. For the former two, it uses the Snyk CLI as a data provider, for the latter it is connecting
+  directly to the Snyk Code API.
+homepage: https://github.com/snyk/snyk-ls
+licenses:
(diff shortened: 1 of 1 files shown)


packages/snyk-ls/package.yaml
chrisgrieser: for backward compatibility, this package should only be marked as deprecated at first and only removed at a later date

Conarius:
We usually have the policy of first marking it as deprecated and once that happens we will remove the package after 6 months

PR #14624, opened 2026-04-17
fix(tombi): use github releases instead of pypi

glmlm:
### Describe your changes
`tombi` is a Rust binary that doesn't require Python. The current PyPI-based installation unnecessarily requires a Python runtime and virtual environment, causing installation failures on systems without `python3-venv`.

This change uses pre-built binaries from GitHub Releases, eliminating the Python dependency.

### DIFF
--- a/packages/tombi/package.yaml
+++ b/packages/tombi/package.yaml
@@ -16,7 +16,29 @@ neovim:
   lspconfig: tombi
 
 source:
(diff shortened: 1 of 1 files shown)


Conarius (approved):
LGTM

chrisgrieser (approved):

PR #14655, opened 2026-04-20
chore(java-test): bump 0.45.0

LDWNS:
### Describe your changes
The java test package has been updated to the latest version.
The fixes are listed here: https://github.com/microsoft/vscode-java-test/releases/0.45.0

The version `0.43.2` is hardcoded in several spots, as per the `package.json` here:
https://github.com/microsoft/vscode-java-test/compare/0.44.0...0.45.0#diff-7ae45ad102eab3b6d7e7896acd08c427a9b25b346470d7bc6507b6481575d519R58
 
### DIFF
--- a/packages/java-test/package.yaml
+++ b/packages/java-test/package.yaml
@@ -26,15 +26,8 @@ source:
   id: pkg:openvsx/vscjava/vscode-java-test@0.45.0
   download:
     file: vscjava.vscode-java-test-{{version}}.vsix
-    plugin_version: 0.43.1
-
-  version_overrides:
-    - constraint: semver:<=0.43.1
-      id: pkg:openvsx/vscjava/vscode-java-test@0.43.1
-      download:
-        file: vscjava.vscode-java-test-{{version}}.vsix
-        plugin_version: "{{version}}"
+    plugin_version: 0.45.0
 
 share:
   java-test/: extension/server/
-  java-test/com.microsoft.java.test.plugin.jar: extension/server/com.microsoft.java.test.plugin-{{ source.download.plugin_version }}.jar
+  java-test/com.microsoft.java.test.plugin.jar: extension/server/com.microsoft.java.test.plugin-0.43.1.jar
--- a/packages/java-test/package.yaml
+++ b/packages/java-test/package.yaml
@@ -23,7 +23,7 @@ categories:
   - DAP
 
 source:
-  id: pkg:openvsx/vscjava/vscode-java-test@0.44.0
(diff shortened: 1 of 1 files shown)


Conarius:
Problem is:
We're sourcing it from OpenVSX. So unless the .vsix package is available in the github repo, we would also need to update the versions manually or via a variable, given that the repo and the OpenVSX are in sync

Conarius (dismissed):
Maybe making it more dynamic by using the `{{ version }}` variable all over the file?

packages/java-test/package.yaml
```diff
@@ -23,18 +23,10 @@ categories:
...
-    - constraint: semver:<=0.43.1
-      id: pkg:openvsx/vscjava/vscode-java-test@0.43.1
-      download:
-        file: vscjava.vscode-java-test-{{version}}.vsix
-        plugin_version: "{{version}}"
 
 share:
   java-test/: extension/server/
-  java-test/com.microsoft.java.test.plugin.jar: extension/server/com.microsoft.java.test.plugin-{{ source.download.plugin_version }}.jar
+  java-test/com.microsoft.java.test.plugin.jar: extension/server/com.microsoft.java.test.plugin-0.43.1.jar
```
Conarius: Maybe use `{{ version }}` here as well, if the .jar file has that version already.

LDWNS:
@Conarius would you mind revisiting this PR?
No pressure, thanks for all the work you do.

packages/java-test/package.yaml
```diff
@@ -23,18 +23,10 @@ categories:
...
   download:
     file: vscjava.vscode-java-test-{{version}}.vsix
-    plugin_version: 0.43.1
-
-  version_overrides:
-    - constraint: semver:<=0.43.1
-      id: pkg:openvsx/vscjava/vscode-java-test@0.43.1
-      download:
-        file: vscjava.vscode-java-test-{{version}}.vsix
-        plugin_version: "{{version}}"
```
williamboman, line 36: I dont see a reason to remove this?

williamboman (commented):
Looks good to me, but I don't see a reason to remove support for older versions?

packages/java-test/package.yaml
LDWNS: Good point. I didn't understand the goal behind this piece of code. I did my due diligence and now understand what it means. Thanks for pointing it out.

williamboman (approved):

Conarius (approved):
LGTM

PR #14660, opened 2026-04-20
Update qmlls version to 0.7

mogwa1:
### Describe your changes

Update to upstream version 0.7 of qmlls.

Tested inside neovim

### DIFF
--- a/packages/qmlls/package.yaml
+++ b/packages/qmlls/package.yaml
@@ -12,16 +12,16 @@ categories:
   - LSP
 
 source:
-  id: pkg:github/TheQtCompanyRnD/qmlls-workflow@0.5
+  id: pkg:github/TheQtCompanyRnD/qmlls-workflow@0.7
   asset:
     - target: [darwin_x64, darwin_arm64]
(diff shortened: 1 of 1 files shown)


Conarius (approved):
LGTM

chrisgrieser (commented):
Is it appropriate to update every single user to nightly?

Most Mason users won't even notice that they switched to a nightly build

mogwa1:
I'm actually not sure why the release files have "nightly" in the name of the package. If you check the other releases of this package, then it's clear that this is a proper versioned release (0.7).  The nightlies use a completely different versioning scheme based on date: "2026.15" etc.

I think the person doing the release just messed up the filenames.
If you really don't trust that this is a proper release, then there is at least an update available to 0.6 that doesn't have "nightly" in the filename.

chrisgrieser:
> I think the person doing the release just messed up the filenames.

I see. Might be worth opening an issue at qmlls then, since having a stable release named "nightly" is quite confusing.

We can merge this in the meantime though.

chrisgrieser (approved):

PR #14717, opened 2026-04-23
bump(java-debug-adapter): update to 0.59.0

aveseli:
### Describe your changes
updates java-debug-adapter to 0.59.0

### DIFF
--- a/packages/java-debug-adapter/package.yaml
+++ b/packages/java-debug-adapter/package.yaml
@@ -10,7 +10,7 @@ categories:
   - DAP
 
 source:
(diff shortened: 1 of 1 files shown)


packages/java-debug-adapter/package.yaml
Conarius: Does this also need to be updated?

aveseli:
I have been testing this update successfully since a few days and it works without problems. would be great if we could have that merged.

williamboman (approved):

Conarius (approved):
LGTM

PR #14806, opened 2026-04-29
fix(kotlin-lsp): bump to version v262.4739.0 and adapt to packaging changes

stonemaster:
### Describe your changes
- bin/intellij-server is used now as a primary binary (kotlin-lsp.sh or kotlin-lsp.cmd has been deprecated)
- download extensions are now platform specific
- packages now include a new kotlin-server-VERSION prefix that will now be stripped.
 
Automatic verison update failed: https://github.com/mason-org/mason-registry/pull/14785

### DIFF
--- a/packages/kotlin-lsp/package.yaml
+++ b/packages/kotlin-lsp/package.yaml
@@ -12,32 +12,32 @@ categories:
 source:
   # renovate:versioning=regex:^kotlin-lsp/v(?<major>\d+)\.(?<minor>\d+)\.(?<patch>\d+)$
   # renovate:datasource=github-releases
-  id: pkg:generic/Kotlin/kotlin-lsp@kotlin-lsp/v262.2310.0
+  id: pkg:generic/Kotlin/kotlin-lsp@kotlin-lsp/v262.4739.0
   download:
     - target: darwin_x64
       files:
-        kotlin-lsp.zip: https://download-cdn.jetbrains.com/kotlin-lsp/{{ version | strip_prefix "kotlin-lsp/v" }}/kotlin-lsp-{{ version | strip_prefix "kotlin-lsp/v" }}-mac-x64.zip
-      bin: kotlin-lsp.sh
+        kotlin-lsp.sit: https://download-cdn.jetbrains.com/kotlin-lsp/{{ version | strip_prefix "kotlin-lsp/v" }}/kotlin-server-{{ version | strip_prefix "kotlin-lsp/v" }}.sit
+      bin: kotlin-server-{{ version | strip_prefix "kotlin-lsp/v" }}/bin/intellij-server
     - target: darwin_arm64
       files:
-        kotlin-lsp.zip: https://download-cdn.jetbrains.com/kotlin-lsp/{{ version | strip_prefix "kotlin-lsp/v" }}/kotlin-lsp-{{ version | strip_prefix "kotlin-lsp/v" }}-mac-aarch64.zip
-      bin: kotlin-lsp.sh
+        kotlin-lsp.sit: https://download-cdn.jetbrains.com/kotlin-lsp/{{ version | strip_prefix "kotlin-lsp/v" }}/kotlin-server-{{ version | strip_prefix "kotlin-lsp/v" }}-aarch64.sit
+      bin: kotlin-server-{{ version | strip_prefix "kotlin-lsp/v" }}/bin/intellij-server
     - target: linux_x64
       files:
-        kotlin-lsp.zip: https://download-cdn.jetbrains.com/kotlin-lsp/{{ version | strip_prefix "kotlin-lsp/v" }}/kotlin-lsp-{{ version | strip_prefix "kotlin-lsp/v" }}-linux-x64.zip
-      bin: kotlin-lsp.sh
+        kotlin-lsp.tar.gz: https://download-cdn.jetbrains.com/kotlin-lsp/{{ version | strip_prefix "kotlin-lsp/v" }}/kotlin-server-{{ version | strip_prefix "kotlin-lsp/v" }}.tar.gz
+      bin: kotlin-server-{{ version | strip_prefix "kotlin-lsp/v" }}/bin/intellij-server
     - target: linux_arm64
       files:
-        kotlin-lsp.zip: https://download-cdn.jetbrains.com/kotlin-lsp/{{ version | strip_prefix "kotlin-lsp/v" }}/kotlin-lsp-{{ version | strip_prefix "kotlin-lsp/v" }}-linux-aarch64.zip
-      bin: kotlin-lsp.sh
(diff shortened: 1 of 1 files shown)


chrisgrieser (approved):

Conarius (approved):
LGTM

packages/kotlin-lsp/package.yaml
```diff
@@ -12,32 +12,32 @@ categories:
...
+      bin: kotlin-server-{{ version | strip_prefix "kotlin-lsp/v" }}/bin/intellij-server.exe
     - target: win_arm64
       files:
-        kotlin-lsp.zip: https://download-cdn.jetbrains.com/kotlin-lsp/{{ version | strip_prefix "kotlin-lsp/v" }}/kotlin-lsp-{{ version | strip_prefix "kotlin-lsp/v" }}-win-aarch64.zip
-      bin: kotlin-lsp.cmd
+        kotlin-lsp.zip: https://download-cdn.jetbrains.com/kotlin-lsp/{{ version | strip_prefix "kotlin-lsp/v" }}/kotlin-server-{{ version | strip_prefix "kotlin-lsp/v" }}-aarch64.win.zip
+      bin: kotlin-server-{{{ version | strip_prefix "kotlin-lsp/v" }}/bin/intellij-server.exe
 
 bin:
   kotlin-lsp: "{{source.download.bin}}"
```
williamboman: I think it'd make sense to also update the `bin` entry here. We don't want to drift away from upstream binary names.

```suggestion
  intellij-server: "{{source.download.bin}}"
```

Also, I think in order to avoid breakage, we could link 2 binaries. Keep the package as-is (i.e. link `kotlin-lsp` like before), but add new entries for `intellij-server` specifically.

When they fully remove the `kotlin-lsp.sh` scripts upstream we'll know about it here because CI will catch it

stonemaster: Thanks for the review and the suggestion!

According to the [latest changelog](https://github.com/Kotlin/kotlin-lsp/releases/tag/kotlin-lsp/v262.4739.0) it seems this is going to be the last release that will contain the old kotlin-lsp.sh script. So the question is whether it makes sense to link two binaries when the next release will make the pipeline fail for sure, and another change is needed to make the automatic update work?

williamboman: > According to the [latest changelog](https://github.com/Kotlin/kotlin-lsp/releases/tag/kotlin-lsp/v262.4739.0) it seems this is going to be the last release that will contain the old kotlin-lsp.sh script. So the question is whether it makes sense to link two binaries when the next release will make the pipeline fail for sure, and another change is needed to make the automatic update work?

Ok I missed that, then it doesn't make sense to include it :+1:

williamboman:
mason.nvim doesn't support `.sit` files, probably why tests are failing on `darwin_*` systems (first time I hear about that file format tbh).

packages/kotlin-lsp/package.yaml
```diff
@@ -12,35 +12,35 @@ categories:
...
       files:
-        kotlin-lsp.zip: https://download-cdn.jetbrains.com/kotlin-lsp/{{ version | strip_prefix "kotlin-lsp/v" }}/kotlin-lsp-{{ version | strip_prefix "kotlin-lsp/v" }}-mac-x64.zip
-      bin: kotlin-lsp.sh
+        kotlin-lsp.sit: https://download-cdn.jetbrains.com/kotlin-lsp/{{ version | strip_prefix "kotlin-lsp/v" }}/kotlin-server-{{ version | strip_prefix "kotlin-lsp/v" }}.sit
+      bin: kotlin-server-{{ version | strip_prefix "kotlin-lsp/v" }}/bin/intellij-server
     - target: darwin_arm64
       files:
-        kotlin-lsp.zip: https://download-cdn.jetbrains.com/kotlin-lsp/{{ version | strip_prefix "kotlin-lsp/v" }}/kotlin-lsp-{{ version | strip_prefix "kotlin-lsp/v" }}-mac-aarch64.zip
-      bin: kotlin-lsp.sh
+        kotlin-lsp.sit: https://download-cdn.jetbrains.com/kotlin-lsp/{{ version | strip_prefix "kotlin-lsp/v" }}/kotlin-server-{{ version | strip_prefix "kotlin-lsp/v" }}-aarch64.sit
```
williamboman: I think the easiest solution is to write it as a `.zip` file, which will enable `mason.nvim` to identify which program to use to unpack it (`unzip`).

`.sit` files are certainly an interesting choice by them and while `unzip` works for these specific files, it fails for pretty much all test files available in https://github.com/ssokolow/stuffit-test-files. I'm not interested in adding proper support for `.sit` in `mason.nvim` so I think this is the best workaround for now.

```suggestion
        kotlin-lsp.zip: https://download-cdn.jetbrains.com/kotlin-lsp/{{ version | strip_prefix "kotlin-lsp/v" }}/kotlin-server-{{ version | strip_prefix "kotlin-lsp/v" }}-aarch64.sit
```

packages/kotlin-lsp/package.yaml
```diff
@@ -12,35 +12,35 @@ categories:
...
   # renovate:versioning=regex:^kotlin-lsp/v(?<major>\d+)\.(?<minor>\d+)\.(?<patch>\d+)$
   # renovate:datasource=github-releases
-  id: pkg:generic/Kotlin/kotlin-lsp@kotlin-lsp/v262.2310.0
+  id: pkg:generic/Kotlin/kotlin-lsp@kotlin-lsp/v262.4739.0
   download:
     - target: darwin_x64
       files:
-        kotlin-lsp.zip: https://download-cdn.jetbrains.com/kotlin-lsp/{{ version | strip_prefix "kotlin-lsp/v" }}/kotlin-lsp-{{ version | strip_prefix "kotlin-lsp/v" }}-mac-x64.zip
-      bin: kotlin-lsp.sh
+        kotlin-lsp.sit: https://download-cdn.jetbrains.com/kotlin-lsp/{{ version | strip_prefix "kotlin-lsp/v" }}/kotlin-server-{{ version | strip_prefix "kotlin-lsp/v" }}.sit
```
williamboman:
```suggestion
        kotlin-lsp.zip: https://download-cdn.jetbrains.com/kotlin-lsp/{{ version | strip_prefix "kotlin-lsp/v" }}/kotlin-server-{{ version | strip_prefix "kotlin-lsp/v" }}.sit
```

stonemaster: Thank you for the suggestion! 

The whole packaging seems to be a little bit non-standard in kotlin-lsp ..

stonemaster:
I also noticed that the windows packaging was broken. Unlike mac and Linux, there is **no** prefix added in windows for yet unknown reasons.

williamboman (approved):

weareflo:
Any ideas about the DAP capabilities they mention in the release log?

Issue #14807, opened 2026-04-29
Add statix nix linter


dolphinau:
Looks like mason have support for https://github.com/oppiliappan/statix

DeusProx:
The PR got merged. Can this issue be closed?

PR #14812, opened 2026-04-29
add statix

dolphinau:
### Describe your changes
Add statix nix linter using cargo

#14807 

The repo has 889 stars on https://github.com/oppiliappan/statix

### Local verification:
- Installed statix from this local file: registry into Neovim.
- Confirmed which statix and vim.fn.exepath('statix') resolved to Mason's installed binary, not the global mise install.
- Ran statix --help from the Mason-managed binary successfully.
- Opened a temporary nix project in headless Neovim and confirmed the statix attached successfully.

### Screenshots
image
image
image
image
image

### DIFF
--- a/packages/statix/package.yaml
+++ b/packages/statix/package.yaml
@@ -5,12 +5,12 @@ homepage: https://github.com/oppiliappan/statix
 licenses:
   - MIT
 languages:
-  - nix
+  - Nix
 categories:
   - Linter
 
 source:
-  id: pkg:cargo/statix@4.8.3
+  id: pkg:cargo/statix@v0.5.8?repository_url=https://github.com/oppiliappan/statix
 
 bin:
   statix: cargo:statix
--- a/packages/statix/package.yaml
+++ b/packages/statix/package.yaml
@@ -1,0 +1,16 @@
+---
+name: statix
+description: lints and suggestions for the nix programming language
+homepage: https://github.com/oppiliappan/statix
+licenses:
+  - MIT
+languages:
+  - nix
+categories:
+  - Linter
+
+source:
+  id: pkg:cargo/statix@4.8.3
+
+bin:
+  statix: cargo:statix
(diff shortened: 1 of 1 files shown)


packages/statix/package.yaml
```diff
@@ -0,0 +1,19 @@
...
+homepage: https://github.com/oppiliappan/statix
+licenses:
+  - MIT
+languages:
+  - nix
+categories:
+  - Linter
+
+source:
+  id: pkg:cargo/statix@4.8.3
```
chrisgrieser: While this is not wrong, it is generally preferable to use github releases over cargo, since not every user has the rust toolchain installed on their system.

dolphinau: It is actually the wrong package... :sweat_smile:

dolphinau: So there are two options:
- use nix but it's not supported in mason
- build from source but it requires the rust toolchain installed

I'll go for the second one so rust toolchain will be required. Don't hesitate if you see a better option

chrisgrieser: ah, just seeing now that they are doing "empty" releases? https://github.com/oppiliappan/statix/releases

Yeah, in that case rust is okay. Though it would be preferrable to open an issue at their repo, asking for builds to be made available.

packages/statix/package.yaml
```diff
@@ -0,0 +1,19 @@
+---
+name: statix
+description: lints and suggestions for the nix programming language
+homepage: https://github.com/oppiliappan/statix
+licenses:
+  - MIT
+languages:
+  - nix
```
chrisgrieser: must be "Nix" (this is why one of the test fails)

dolphinau:
New source successfully installed
This time it's attaching and reporting successfully

image
image


image

packages/statix/package.yaml
dolphinau: resolved in last commit

chrisgrieser (approved):

williamboman (approved):

PR #15056, opened 2026-05-16
security: add minimumReleaseAge for all datasources

williamboman:
This adds a 1 day default minimum release age for all datasources and extending that to 3 days for npm packages for
reasons listed in the comment. Since we have Renovate's automerge config enabled, Renovate will still immediately create
PRs as soon as it detects version upgrade but will wait with automerge until the mandated minimumReleaseAge has elapsed.

### DIFF
--- a/renovate.json5
+++ b/renovate.json5
@@ -1,6 +1,19 @@
 {
     extends: ["github>mason-org/registry-renovate-config"],
+    // Order of rules matter, later rules override (merge) previous ones.
     packageRules: [
+        {
+            "matchDatasources": ["*"],
+            "minimumReleaseAge": "1 day"
+        },
+        {
+            /*
+            * npm packages less than 72 hours (3 days) old can be unpublished from the npm registry, which could result
+            * in a service impact if they have already been upgraded.
+            */
+            "matchDatasources": ["npm"],
+            "minimumReleaseAge": "3 days"
+        },
         {
             matchDepNames: ["cucumber-language-server"],
             allowedVersions: "!/^1\\.[34]\\.0/"


Conarius (approved):
LGTM

chrisgrieser:
I like the general idea, but I wonder if this could/should maybe be configurable on the user side?

Like for instance, when I trust the dev of a package and need to give feedback to some changes right away, this adds a one day delay or requires me to install a package manually. 

In the other side, there might be a user who does not mind delayed updates at all and prefers extra security, so that user might want to increase the delay to one week.

williamboman:
> I like the general idea, but I wonder if this could/should maybe be configurable on the user side?

That's an interesting idea, but a bit tricky/impossible to implement without other side effects due to how the registry functions. The only approach I can think of spontaneously would be to pull the latest registry release that meets whatever "minimumReleaseAge" the user has configured, but that would require paging the GitHub releases API and it'd also mean any changes to the registry that are not related to version upgrades would be excluded (new packages, other fixes, etc.). There might also be cases where a new version upgrade is urgent because older versions are vulnerable, if the minimumReleaseAge is enforced client-side I can't see a feasible way to version bump such packages immediately.

I don't think the average user would notice the delay at all except for maybe a very small set of packages that have a fixed release cadence (e.g. rust-analyzer who cuts a new release every Monday). In case a popular package receives a new big version upgrade and people notice it's not available yet, we can always manually merge Renovate PRs ahead of time.

chrisgrieser:
I see, yeah if it's more trouble than it's worth, then the current setup is totally fine

chrisgrieser (approved):

williamboman:
Let's give this a try and see how we like it

PR #15136, opened 2026-05-22
feat(new-package): kulala-ls

rafael-luigi-bekkema:
Refs mason-org/mason.nvim#1806

### Describe your changes
Added the kulala-ls package.

[mason-org/mason.nvim#1806](https://github.com/mason-org/mason.nvim/issues/1806)

https://github.com/neovim/nvim-lspconfig/blob/master/lsp/kulala_ls.lua

### DIFF
--- a/packages/kulala-ls/package.yaml
+++ b/packages/kulala-ls/package.yaml
@@ -1,0 +1,13 @@
+---
+name: kulala-ls
+description: Language server for kulala / HTTP files
+homepage: https://github.com/mistweaverco/kulala-ls
+licenses: [MIT]
+languages: [http]
+categories: [LSP]
+source:
+  id: pkg:npm/%40mistweaverco/kulala-ls@1.11.1
+bin:
+  kulala-ls: npm:kulala-ls
+neovim:
+  lspconfig: kulala_ls


packages/kulala-ls/package.yaml
```diff
@@ -0,0 +1,12 @@
+name: kulala-ls
```
williamboman, line 2:
```suggestion
---
name: kulala-ls
```

williamboman (commented):
Needs a YAML document separator at the top, but otherwise LGTM

williamboman (approved):
LGTM

gorillamoe:
kulala-ls could be still used in theory, but it has been super-seeded by Kulala-Core, which is the all-in-one backend that powers kulala.nvim, kulala.vscode and kulala-fmt (standalone desktop coming soon) now. You can use it with other tools as well. It includes the LSP and everything else that you need to run requests from .http files (it does all the heavy lifting of parsing and running requests and js/lua/ts-scripts and reading from different sources like your process.env, config files etc.).

Since the kulala family tools auto-download their required core version by themselves, adding it to the mason registry would only make sense, if you want to use the LSP as a standalone with something else than kulala.

I'm not sure there are other tools outside the Kulala family working with Kulala-Core at the moment.

I know that there is a Zed implementation based off Kulala, but their implementation is still using the older 5x range and probably also has to migrate at some point.

rafael-bekkema-alserdamedia:
ok, let's close this then

PR #15704, opened 2026-06-25
Update shfmt to include Zsh support

### LINKED ISSUE #15703: Add zsh to shfmt
It looks like back in March, [shfmt](https://github.com/mvdan/sh) [clarified that they support zsh in their README](https://github.com/mvdan/sh/commit/8e406a63846b8e150988d7b4008739727c268d60). Not sure when the code changes for Zsh support actually went in though. But regardless, the provider in the registry only lists bash, mksh, and shell. So zsh needs to be added to the file list for shfmt.

brdoney:
### Describe your changes
Just a small fix to enable Zsh support in shfmt, which was added at some point (and recently officially added to their README)

Fixes #15703

### DIFF
--- a/packages/shfmt/package.yaml
+++ b/packages/shfmt/package.yaml
@@ -1,6 +1,6 @@
 ---
 name: shfmt
-description: A shell formatter (sh/bash/mksh).
+description: A shell formatter (sh/bash/zsh/mksh).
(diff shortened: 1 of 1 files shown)


chrisgrieser (approved):

Conarius (approved):
LGTM

Issue #15934, opened 2026-07-09
vue-language-server crashes on startup because it bundles incompatible TypeScript 7.0+


s-montigny-desautels:
**Describe the bug**
When installing `vue-language-server` via Mason, it currently pulls down the `latest` version of `typescript` as a dependency (which is now `7.0.x`). `vue-language-server` is not yet compatible with the latest typescript version, that is now written in Go.

When `vue-language-server` starts up and attempts to use this bundled TypeScript version to parse files, it instantly crashes with a `TypeError`.

**To Reproduce**

1. Install or update `vue-language-server` via Mason (`:MasonInstall vue-language-server`).
2. Open a `.vue` file in Neovim.
3. The server crashes immediately.

**Expected behavior**
The language server should start up successfully and attach to the buffer. `Mason` should pin the typescript version that is included in the package to version `6.x.x`.

**Error Log**

```text
[ERROR] "rpc" "vue-language-server" "stderr" 
"/home/<user>/.local/share/nvim/mason/packages/vue-language-server/node_modules/@vue/language-server/lib/server.js:40
                        projectInfoPromise = sendTsServerRequest('_vue:' + ts.server.protocol.CommandTypes.ProjectInfo, {
                                                                                             ^

TypeError: Cannot read properties of undefined (reading 'protocol')
    at Object.getLanguageService (/home/<user>/.local/share/nvim/mason/packages/vue-language-server/node_modules/@vue/language-server/lib/server.js:40:86)
    ...

```

**Additional Context**
By looking at the `vue_ls` source code, it resolves TS via `require('typescript')` if the `--tsdk` flag isn't provided. Because Mason installs TS 7.0 alongside it, it loads the incompatible TS engine and fails at `(0, server_1.startServer)(ts);`. Manually forcing the LSP to boot using an older TypeScript engine (e.g., via `vtsls`'s bundled TS version) prevents the crash.

```lua
   local tsdk = vim.fn.stdpath("data")
        .. "/mason/packages/vtsls/node_modules/@vtsls/language-server/node_modules/typescript/lib"
    vim.lsp.config("vue_ls", {
        cmd = { "vue-language-server", "--stdio", "--tsdk=" .. tsdk },
    })
```

hanool:
Related to #15985.

TL;DR workaround

```sh
cd ~/.local/share/nvim/mason/packages/vue-language-server
npm install typescript@6.0.3 --no-save
```

[...]

PR #16029, opened 2026-07-15
fix(vuels): pin typescript to 6.0.3, since 7.x is not compatible with vue yet

s-montigny-desautels:
### Describe your changes
Typescript 7.x is not compatible with the vue language server.

#15934 

### DIFF
--- a/packages/vue-language-server/package.yaml
+++ b/packages/vue-language-server/package.yaml
@@ -12,6 +12,7 @@ categories:
 source:
   id: pkg:npm/%40vue/language-server@3.3.7
   extra_packages:
+    # vue_ls doesn't support typescript v7, yet. See https://github.com/mason-org/mason-registry/issues/15934
     - "typescript@6.0.3"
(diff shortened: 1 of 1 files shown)


Conarius (approved):
LGTM

packages/vue-language-server/package.yaml
```diff
@@ -12,7 +12,7 @@ categories:
 source:
   id: pkg:npm/%40vue/language-server@3.3.7
   extra_packages:
-    - typescript
+    - "typescript@6.0.3"
```
chrisgrieser, line 16: just for the sake of transparency, I'd add a comment with a link to issue #15934, so it's clear why the typescript version is pinned. Otherwise looking good to me.

s-montigny-desautels: Good idea, I've added a comment linking to the issue. Thanks !

chrisgrieser (approved):

PR #16355, opened 2026-08-05
fix(ruff): use github releases instead of pypi

glmlm:
### Describe your changes

`ruff` is a Python-independent binary. The current installation method requires a Python runtime, leading to failures on systems without `python3-venv`.

This change switches to pre-built binaries from GitHub Releases, eliminating the Python dependency.


chrisgrieser (approved):

Conarius (approved):
LGTM

PR #16619, opened 2026-08-24
feat: add ms-terraform-lsp

tobiassen:
### Describe your changes

Adds the Microsoft Terraform Providers Language Server, which provides completion, hover documentation and schema validation for the `azapi`, `azurerm` and `msgraph` Terraform providers, and for Azure Verified Modules.

Linux targets are plain rather than `_gnu` because these are static Go builds, matching the existing terraform-ls package.

One thing worth flagging, the binary is around 426 MB unpacked, because the Azure resource schemas are embedded in it.

Approved at nvim-lspconfig: neovim/nvim-lspconfig#4510, merged as `ms_terraform_lsp`

      Installed on darwin_arm64 and checked the asset templating with `--target=linux_x64` and `--target=win_x64`. All three resolve the expected release asset and unpack a single binary at the archive root. After installing, the server attaches in Neovim alongside terraformls and completes property names inside an `azapi_resource` body.

### DIFF
--- a/packages/ms-terraform-lsp/package.yaml
+++ b/packages/ms-terraform-lsp/package.yaml
@@ -1,0 +1,46 @@
+---
+name: ms-terraform-lsp
+description: |
+  Microsoft Terraform Providers Language Server. Completion, hover documentation and schema validation for the azapi,
+  azurerm and msgraph Terraform providers, and for Azure Verified Modules.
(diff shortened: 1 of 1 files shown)


Conarius (approved):
LGTM

chrisgrieser (approved):

PR #16663, opened 2026-08-27
feat(package): add cx

eptx:
Adds **CX** (https://cxhome.org) — the data and code language. One binary carries the CLI, the language server (`cx lsp`), and libcx.

Release assets are flat, stable-named tarballs (`cx-darwin-arm64.tar.gz`, `cx-linux-arm64.tar.gz`) with the `cx` binary at the tar root, published per release at https://github.com/cx-home/cx/releases (current: v0.17.0). Further targets (linux x86_64, windows) join as the release lane ships them.

Verified locally: both assets download and extract, `cx --version` reports the release headline, `cx lsp --stdio` speaks LSP.

### DIFF
--- a/packages/cx/package.yaml
+++ b/packages/cx/package.yaml
@@ -1,0 +1,23 @@
+name: cx
+description: CX — the data and code language. One binary carries the CLI, the language server (cx lsp), and libcx.
+homepage: https://cxhome.org
+licenses:
+  - Apache-2.0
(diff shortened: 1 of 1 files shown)


chrisgrieser (commented):
Hi, unfortunately the package did not fulfill the requirements for new packages yet. (See the link sieh when you open a new PR)

